PR #1, opened 2017-06-29
Client for release and channel management

areed:
Generate models from the Vendor API's Swagger spec. The generated http transport code was too messy to be useful.

The tests are integration tests, not unit tests. Perhaps they should be in a separate package?

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -6,17 +6,38 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+
+	channels "github.com/replicatedhq/replicated/gen/go/channels"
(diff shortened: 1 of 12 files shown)


client/client.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+// A Client communicates with the Replicated Vendor API.
+type Client struct {
```
emosbaugh: wanna add a Client interface?

areed: added

PR #11, opened 2017-07-14
install script for linux

areed:
README example for CI


README.md
```diff
@@ -1,19 +1,93 @@
...
+* ```REPLICATED_APP_SLUG```
+* ```REPLICATED_API_TOKEN```
+
+Then the above command would be simply
+```
+replicated channel ls
+```
+
+### CI Example
+Creating a new release for every tagged build is a common use of the ```replicated``` command.
```
emosbaugh: are you sure about the triple tildes inline?

PR #13, opened 2017-07-23
Improve UX for vendor CLI

areed:
Specify app as either slug or ID
Implement channel adoption, counts, and releases with tests.
Remove adoption, counts, and releases section from channel inspect.
Add release sequence and release version label to channel inspect.
Align columns in channel inspect and release inspect.
Add CreateLicense method to client using v2 swagger spec.
Delete apps in single step to avoid rate limits in integration tests.
Remove version from release ls becuase release version is unused and
easily confused with channel release version.
Sort channels by position and channel releases newest to oldest.
Don't show help subcommand in available commands.


client/client.go
```diff
@@ -5,18 +5,20 @@ import (
...
+	v2 "github.com/replicatedhq/replicated/gen/go/v2"
 )
 
 const apiOrigin = "https://api.replicated.com/vendor"
 
 // Client provides methods to manage apps, channels, and releases.
 type Client interface {
-	GetAppBySlug(slug string) (*apps.App, error)
+	GetApp(slugOrID string) (*apps.App, error)
 	CreateApp(name string) (*apps.App, error)
```
emosbaugh: i kinda prefer something like this for forwards compatibility. Probably applies to other function signatures as well

```go
CreateApp(opts CreateAppOptions) (*apps.App, error)
```

cli/cmd/channel_adoption.go
```diff
@@ -0,0 +1,36 @@
...
+	"errors"
+
+	"github.com/replicatedhq/replicated/cli/print"
+	"github.com/spf13/cobra"
+)
+
+var channelAdoptionCmd = &cobra.Command{
+	Use:   "adoption CHANNEL_ID",
+	Short: "Print channel adoption statistics by license type",
+	Long:  "Print channel adoptions statistics by license type",
```
emosbaugh: typo

areed:
@emosbaugh How's this? Identifiers as arguments and options as struct. The options type can be used in either create or update calls.

emosbaugh:
lgtm

PR #19, opened 2017-09-17
Allow reading yaml from stdin

kichik:
This is useful on Windows where the "$(< replicated.yml)" syntax is not supported.
It allows for `type replicated.yml | replicated release create create --promote Unstable --yaml -`


areed:
Thanks @kichik - this is great. Let me add a commit to update the tests and then I'll merge.

kichik:
Thanks!

PR #27, opened 2018-12-07
add release-promote flags to release-create --promote

dexhorthy:
reopened version of https://github.com/replicatedhq/replicated/pull/26 on upstream to see if that gets the tests passing


client/release.go
```diff
@@ -21,7 +21,7 @@ func (c *HTTPClient) ListReleases(appID string) ([]releases.AppReleaseInfo, erro
 // CreateRelease adds a release to an app.
 func (c *HTTPClient) CreateRelease(appID string, opts *ReleaseOptions) (*releases.AppReleaseInfo, error) {
 	path := fmt.Sprintf("/v1/app/%s/release", appID)
-	body := &releases.Body{
+	body := &releases.BodyCreateRelease{
```
dexhorthy: I renamed a couple of these where `Body` was actually something like `Body5` in the generated mono-package. There are still some `BodyXX` lying around but they are unused for now

README.md
```diff
@@ -141,6 +141,14 @@ Releases are created on Travis when a tag is pushed. This will also update the d
 git tag -a v0.1.0 -m "First release" && git push upstream v0.1.0
 ```
 
+### Regenerating Client Code
+
+When the swagger definitions change, you can regenerate the Client code from the swagger spec with
+
+    make get-spec-prod gen-models
+
+models for the v2 api isn't really working yet, need to find the URL for that OpenAPI spec.
```
dexhorthy: we might even consider removing the v2 api scripts since we don't guarantee that those are stable

PR #30, opened 2019-03-12
Ship and Platform clients

marccampbell:


README.md
```diff
@@ -101,14 +101,14 @@ import (
 	"log"
 	"os"
 
-	"github.com/replicatedhq/replicated/client"
+	"github.com/replicatedhq/replicated/pkg/platform_client"
```
dexhorthy: I think underscores are discouraged in package names https://blog.golang.org/package-names

cli/cmd/root.go
```diff
@@ -102,18 +106,38 @@ func Execute(stdin io.Reader, stdout io.Writer, stderr io.Writer) error {
...
 
-		app, err := api.GetApp(appSlugOrID)
+		appType, err := commonAPI.GetAppType(appSlugOrID)
 		if err != nil {
 			return err
 		}
-		runCmds.appID = app.Id
+		runCmds.appType = appType
+
+		if appType == "platform" {
```
dexhorthy: yeah this feels like the kind of thing that would benefit from an interface but I don't really see a clear path to getting there yet

cli/cmd/root.go
```diff
@@ -20,15 +23,16 @@ const (
...
+var shipOrigin = "https://g.replicated.com/graphql"
 
 func init() {
 	RootCmd.PersistentFlags().StringVar(&appSlugOrID, "app", "", "The app slug or app id to use in all calls")
 	RootCmd.PersistentFlags().StringVar(&apiToken, "token", "", "The API token to use to access your app in the Vendor API")
 
 	originFromEnv := os.Getenv("REPLICATED_API_ORIGIN")
 	if originFromEnv != "" {
-		apiOrigin = originFromEnv
+		platformOrigin = originFromEnv
```
emosbaugh: do you wanna add something similar to here for ship? so we can test with staging?

client/app.go
```diff
@@ -1,62 +1,59 @@
...
-	resp, err := http.DefaultClient.Do(req)
-	if err != nil {
-		return fmt.Errorf("DeleteApp (%s %s): %v", req.Method, endpoint, err)
-	}
-	resp.Body.Close()
-	if resp.StatusCode != http.StatusNoContent {
-		return fmt.Errorf("DeleteApp (%s %s): status %d", req.Method, endpoint, resp.StatusCode)
-	}
+func (c *Client) CreateApp(appOptions interface{}) (interface{}, error) {
+	return nil, nil
```
emosbaugh: i assume you will fill these in before merging?

pkg/ship_client/client.go
```diff
@@ -0,0 +1,129 @@
...
+	body, err := json.Marshal(requestObj)
+	if err != nil {
+		return errors.Wrap(err, "marshal body")
+	}
+
+	bodyReader := bytes.NewReader(body)
+
+	req, err := http.NewRequest("POST", c.GQLServer.String(), bodyReader)
+	if err != nil {
+		return errors.Wrap(err, "marshal body")
```
emosbaugh: wrong message

pkg/ship_client/client.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", c.Token)
+
+	client := http.DefaultClient
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return errors.Wrap(err, "marshal body")
```
emosbaugh: wrong message

pkg/ship_client/client.go
```diff
@@ -0,0 +1,129 @@
...
+		return errors.New("nil response from gql")
+	}
+
+	if resp.Body == nil {
+		return errors.New("nil response.Body from gql")
+	}
+
+	responseBody, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return errors.Wrap(err, "marshal body")
```
emosbaugh: wrong message

pkg/ship_client/app.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (c *GraphQLClient) GetApp(appID string) (*types.App, error) {
+	apps, err := c.ListApps()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, app := range apps {
+		if app.App.ID == appID || app.App.Slug == appID {
```
emosbaugh: seems like there should be a better way

client/client.go
```diff
@@ -1,115 +1,44 @@
...
-	return c
+	return client
 }
 
-func NewHTTPClient(origin string, apiKey string) Client {
-	c := &HTTPClient{
-		apiKey:    apiKey,
-		apiOrigin: origin,
+func (c *Client) GetAppType(appID string) (string, error) {
+	platformApps, err := c.PlatformClient.ListApps()
```
emosbaugh: GetApp() ?

cli/cmd/root.go
marccampbell: agreed.. i'm going to add some more implementation to ship methods, and look for opportunity for a pattern to emerge

client/app.go
marccampbell: Yes, but not a blocker here. The CreateApp function in the CLI still calls the platformapi method directly. My intent isn't to add all ship functionality in 1 PR, but to start adding it, as needed.

PR #31, opened 2019-03-14
Linting for ship apps

marccampbell:
This adds a new command:

`replicated release lint --yaml "$(cat ./ship.yaml)"`

If the current app (`REPLICATED_APP`) is a ship app, it will use the API token to send the YAML to the GraphQL endpoint specified, and lint, returning errors.

Linting on the server side is: a) easier for now (the linter is JS, this project is go).  and b) allows for linting of dynamic dependencies (github assets, registry credentials, etc)

### DIFF
--- a/cli/cmd/release_lint.go
+++ b/cli/cmd/release_lint.go
@@ -1,0 +1,48 @@
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	"github.com/replicatedhq/replicated/cli/print"
+	"github.com/spf13/cobra"
+)
+
+var lintReleaseYaml string
+
+var releaseLintCmd = &cobra.Command{
+	Use:   "lint",
+	Short: "Lint a YAML",
+	Long:  "Lint a YAML",
+}
+
+func init() {
+	releaseCmd.AddCommand(releaseLintCmd)
+
+	releaseLintCmd.Flags().StringVar(&lintReleaseYaml, "yaml", "", "The YAML config to lint. Use '-' to read from stdin")
+}
+
+func (r *runners) releaseLint(cmd *cobra.Command, args []string) error {
+	if lintReleaseYaml == "" {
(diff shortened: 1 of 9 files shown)


marccampbell:
Note: this is not yet implemented for platform apps

cli/cmd/release_lint.go
```diff
@@ -0,0 +1,48 @@
...
+
+	releaseLintCmd.Flags().StringVar(&lintReleaseYaml, "yaml", "", "The YAML config to lint. Use '-' to read from stdin")
+}
+
+func (r *runners) releaseLint(cmd *cobra.Command, args []string) error {
+	if lintReleaseYaml == "" {
+		return fmt.Errorf("yaml is required")
+	}
+
+	if lintReleaseYaml == "-" {
```
emosbaugh: it looks like stdin is required? do you want to make it support a file path or perhaps ditch the flag? or am i missing something?

pkg/platformclient/release.go
```diff
@@ -84,3 +85,7 @@ func (c *HTTPClient) PromoteRelease(appID string, sequence int64, label, notes s
 	}
 	return nil
 }
+
+func (c *HTTPClient) LintRelease(appID string, yaml string) ([]types.LintMessage, error) {
+	return nil, nil
```
emosbaugh: seems appropriate to return the error here

cli/cmd/release_lint.go
marccampbell: This is the same behavior as "CreateRelease". I would like to support a file path! But thoughts on keeping this consistent for now, and handling that separately?

emosbaugh: sure. the flag just seems weird but im all for consistency

pkg/platformclient/release.go
marccampbell: done

PR #34, opened 2019-03-22
add entitlements client

dexhorthy:
mostly copied from https://github.com/replicatedhq/replicated-starter-ship/tree/entitlements


laverya:
I feel like we should be moving this to Circle at some point
Also moving to Go 1.12 instead of 1.8
Not the PR for that though

cli/cmd/entitlements_set_value.go
```diff
@@ -0,0 +1,108 @@
...
+			}
+
+			bytes, _ := json.MarshalIndent(created, "", "  ")
+			fmt.Printf("%s\n", bytes)
+
+			return nil
+		},
+	}
+
+	cmd.Flags().String("replicated-api-token", "", "Token to use to communicate with https://g.replicated.com")
```
laverya: why is this (and `replicated-app`, and `replicated-api-server`, and possibly others) a per-subcommand flag and not one for the parent command?

dexhorthy: ah good catch, that was a relic of when this was a standalone command tree. Will remove

dexhorthy: This turned into kind of a mess. Since this is somewhat temporary I'm considering just leaving this how it is.

cli/cmd/entitlements_getcustomerrelease.go
```diff
@@ -0,0 +1,74 @@
...
+	Short: "Preview the end customer release view",
+	Long: `Preview the end customer release view for a given customer ID + installation ID`,
+}
+
+func init() {
+	entitlementsCmd.AddCommand(entitlementsGetCustomerRelease)
+
+	entitlementsGetCustomerRelease.Flags().StringVar(&entitlementsGetReleaseCustomerID,"customer-id", "", "customer id")
+	entitlementsGetCustomerRelease.Flags().StringVar(&entitlementsGetReleaseInstallationID,"installation-id", "", "installation id")
+	entitlementsGetCustomerRelease.Flags().StringVar(&entitlementsGetReleaseAPIServer,"replicated-api-server", "https://pg.replicated.com/graphql", "Upstream api server")
```
laverya: I think this one is a duplicate?

dexhorthy: almost -- this changes the previously-defined flag to default to `pg.` intead of `g.`. This command is a special case which simulates the customer view.

laverya: Gotcha, hadn't noticed that
Might be a good comment to add

laverya:
```suggestion
	// change the default from g.replicated.com to pg.replicated.com
	entitlementsGetCustomerRelease.Flags().StringVar(&entitlementsGetReleaseAPIServer,"replicated-api-server", "https://pg.replicated.com/graphql", "Upstream api server")
```

PR #37, opened 2019-04-30
feat: wrap main body in function

josegonzalez:
Wrapping this in a function allows safe usage of `curl | bash`, as it ensures that the script is not executed until the _entire_ script is downloaded.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -21,22 +21,27 @@ download_tar() {
 
 	curl --silent --location --fail "$URL"
 }
-default_dir=/usr/local/bin
-tar_binary=replicated
 
-if [[ -z "$replicated_bindir" ]]; then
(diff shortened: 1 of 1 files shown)


dexhorthy:
Hi @josegonzalez ! this looks great. I believe this script is generated by goreleaser, but I'm not 100% on that. Either way, I don't see any issue with merging this in.

Thanks!

josegonzalez:
Based on [commit history](https://github.com/replicatedhq/replicated/commit/d6901529770cb119071adf3e8fcea9de00f2efbd#diff-3fbb47e318cd8802bd325e7da9aaabe8), I don't think goreleaser generates it (I also tried to find it in goreleaser source, but didn't).

In any case, feel free to merge at your leisure :)

PR #38, opened 2019-05-01
Pact consumer

marccampbell:


.travis.yml
```diff
@@ -4,8 +4,10 @@ services:
 language: go
 go: 1.11.6
```
laverya:
```suggestion
 go:
   - "1.10"
   - "1.11"
   - "1.12"
   - master
 matrix:
   allow_failures:
   - go: master
```

laverya: This is probably unneeded at this time
But we may wish to update the golang version soonish

PR #47, opened 2019-05-24
use a better versioning system for pact tests

laverya:
This way we keep the 'best' of both worlds - we don't need to release this in order to publish the new pact tests, but we avoid cluttering up the pact test matrix with every commit that gets pushed.

Tags will have versions like `v0.10.0`, untagged versions like `v0.10.0-postrelease`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,8 @@ script:
   - make pacts
   - make test
   - make build
-  - make publish-pacts
+#  - make publish-pacts-prerelease
 after_success:
   - test -n "$TRAVIS_TAG" && make package_docker_docs
   - test -n "$TRAVIS_TAG" && curl -sL https://git.io/goreleaser | bash
+  - test -n "$TRAVIS_TAG" && make publish-pacts
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ script:
   - make pacts
   - make test
   - make build
-#  - make publish-pacts-prerelease
 after_success:
   - test -n "$TRAVIS_TAG" && make package_docker_docs
   - test -n "$TRAVIS_TAG" && curl -sL https://git.io/goreleaser | bash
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -36,6 +35,15 @@ publish-pacts:
...
 		https://replicated-pact-broker.herokuapp.com/pacts/provider/vendor-graphql-api/consumer/replicated-cli/version/$(VERSION)
 
+publish-pacts-postrelease:
+	curl \
+		--silent --output /dev/null --show-error --fail \
+		--user ${PACT_BROKER_USERNAME}:${PACT_BROKER_PASSWORD} \
+		-X PUT \
+		-H "Content-Type: application/json" \
+		-d@pacts/replicated-cli-vendor-graphql-api.json \
+		https://replicated-pact-broker.herokuapp.com/pacts/provider/vendor-graphql-api/consumer/replicated-cli/version/$(VERSION)-postrelease
```
marccampbell: @laverya Let's use the "pre-release" syntax instead here. Instead of 0.10.0 post-release, Use 0.11.0-alpha. This is more consistent with our naming.

I don't know how to figure out the next version, maybe we can (for now) just assume minor releases. Thinking through a patch release strategy here, i don't think this creates a big problem.

.travis.yml
```diff
@@ -7,7 +7,8 @@ script:
   - make pacts
   - make test
   - make build
-  - make publish-pacts
+  - make publish-pacts-postrelease
```
marccampbell: "prerelease" (see other comment for more info)

Makefile
laverya: Calculating the next semver is... rather nontrivial in bash
I can add it, but it'll be a fair bit of extra code

laverya: Calculating the next minor version _is_ relatively trivial though
So I can do that

laverya:
This has since been updated to only release pacts on tag push

PR #49, opened 2019-06-14
troubleshoot collector addition to Vendor CLI

smithers1221:


smithers1221:
[ch20135]

pkg/platformclient/collector.go
```diff
@@ -0,0 +1,354 @@
...
+
+type PlatformChannel struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+func (c *HTTPClient) ListCollectors(appID string, appType string) ([]v1.AppCollectorInfo, error) {
+	response := GraphQLResponseListCollectors{}
+
+	request := graphql.Request{
```
divolgin: There shouldn't be any GQL requests in this package.

client/collector.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/replicatedhq/replicated/pkg/types"
+)
+
+func (c *Client) ListCollectors(appID string, appType string) ([]types.CollectorInfo, error) {
+	appType, err := c.GetAppType(appID)
+	if err != nil {
+		return nil, err
+	}
+
+	if appType == "platform" {
```
divolgin: both `ListCollectors` functions are identical. You don't need this split logic here.

client/collector.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+
+		return shipCollectorInfos, nil
+	}
+
+	return nil, errors.New("unknown app type")
+}
+
+func (c *Client) UpdateCollector(appID string, appType string, specID string, yaml string) (interface{}, error) {
+	if appType == "platform" {
```
divolgin: both `UpdateCollector` functions are identical.

client/collector.go
```diff
@@ -0,0 +1,130 @@
...
+		return c.ShipClient.GetCollector(appID, appType, specID)
+	}
+
+	return nil, errors.New("unknown app type")
+
+}
+
+func (c *Client) PromoteCollector(appID string, appType string, specID string, channelIDs ...string) error {
+
+	if appType == "platform" {
```
divolgin: This is the only functions that needs to make distinction between platform and ship apps

PR #52, opened 2019-07-02
version cmd redux

laverya:
this allows `replicated version` to return results

For instance:
```
{
    "version": "v0.10.0-27-gf526188c",
    "git": "f526188",
    "buildTime": "2019-07-02T23:05:52Z",
    "go": {
        "version": "go1.12.6",
        "compiler": "gc",
        "os": "darwin",
        "arch": "amd64"
    }
}
```

`version` is derived from `git describe`, so releases will properly respond with `v0.11.0` or whatever the current version happens to be. Prerelease tests will receive the `<previous-tag>-<commits-since-tag>-<previous-commit-sha>` format.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,0 +1,70 @@
+package version
+
+import (
+	"runtime"
+	"time"
+)
+
+var (
+	build Build
+)
+
+// Build holds details about this build of the Ship binary
+type Build struct {
+	Version      string    `json:"version,omitempty"`
+	GitSHA       string    `json:"git,omitempty"`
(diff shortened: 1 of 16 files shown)


pkg/version/version.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"runtime"
+	"time"
+)
+
+var (
+	build Build
+)
+
+// Build holds details about this build of the Ship binary
```
areed: Ship

laverya: ouch

PR #55, opened 2019-07-24
troubleshoot/collector cli

smithers1221:


pkg/shipclient/collector.go
```diff
@@ -337,19 +332,14 @@ mutation createSupportBundleSpec($name: String, $appId: String, $spec: String, $
 
 	request = graphql.Request{
 		Query: `
-mutation updateSupportBundleSpec($id: ID!, $spec: String!, $githubRef: GitHubRefInput, $isArchived: Boolean) {
```
emosbaugh: this should not be an arg

https://github.com/replicatedhq/vandoor/blob/master/vendor-web/src/mutations/SupportBundleMutations.js#L23

PR #62, opened 2019-09-16
channel functionality for kots apps

smithers1221:


pkg/kotsclient/channel.go
```diff
@@ -1,7 +1,141 @@
...
+			"channelName": name,
+			"description": description,
+		},
+	}
+
+	if err := c.ExecuteRequest(request, &response); err != nil {
+		return err
+	}
+
+	if len(response.Errors) != 0 {
```
emosbaugh: i think this is already handled for you here https://github.com/replicatedhq/replicated/blob/master/pkg/graphql/client.go#L90-L95

PR #67, opened 2019-10-07
adding channel inspect for ship and kots, release update for ship

smithers1221:


client/channel.go
```diff
@@ -41,12 +42,40 @@ func (c *Client) ListChannels(appID string) ([]types.Channel, error) {
...
+	appType, err := c.GetAppType(appID)
+	if err != nil {
+		return err
+	}
+
+	if appType == "platform" {
+		return c.PlatformClient.ArchiveChannel(appID, channelID)
+	} else if appType == "ship" {
+		// return c.ShipClient.ArchiveChannel(appID, channelID)
+		return nil
```
dexhorthy: do you think if this finishes without an error, that the user will believe that the channel was archived?

pkg/kotsclient/channel.go
```diff
@@ -130,6 +140,80 @@ func ArchiveChannel(appID string, channelID string) error {
...
+      repoFullName
+      branch
+      path
+    }
+    extraLintRules
+    created
+    updated
+    isDefault
+    isArchived
+    releases {
```
dexhorthy: do we need to fetch every release on the channel?

pkg/shipclient/channel.go
```diff
@@ -121,6 +131,82 @@ func ArchiveChannel(appID string, channelID string) error {
...
+        repoFullName
+        branch
+        path
+      }
+      extraLintRules
+      created
+      updated
+      isDefault
+      isArchived
+      releases {
```
dexhorthy: same as above, do for fetching channels, do we need to get every release that was published to the channel?

client/channel.go
smithers1221: good point - will add that to the 'sorry, come back later' list.

pkg/kotsclient/channel.go
smithers1221: @dex - see your point, but can we table that for another PR.  Currently pulling from what's on Vendorweb.

PR #70, opened 2019-10-30
kondo-ing the cli

smithers1221:


client/collector.go
```diff
@@ -78,16 +68,11 @@ func (c *Client) GetCollector(appID string, specID string) (*collectors.AppColle
...
-		return err
-	}
+func (c *Client) PromoteCollector(appID string, appType string, specID string, channelIDs ...string) error {
 
 	if appType == "platform" {
 		return c.PlatformClient.PromoteCollector(appID, specID, channelIDs...)
-	} else {
+	} else if appType == "ship" {
 		return c.ShipClient.PromoteCollector(appID, specID, channelIDs...)
 	}
```
fshot: Is it intentional that we're not specifically explaining that Kots apps collectors are handled differently, as we do with the CreateCollector call above?

client/entitlements.go
```diff
@@ -6,11 +6,7 @@ import (
 	"github.com/replicatedhq/replicated/pkg/types"
 )
 
-func (c *Client) CreateEntitlementSpec(name string, spec string, appID string) (*types.EntitlementSpec, error) {
-	appType, err := c.GetAppType(appID)
-	if err != nil {
-		return nil, err
-	}
+func (c *Client) CreateEntitlementSpec(name string, appType string, spec string, appID string) (*types.EntitlementSpec, error) {
```
fshot: It's a little weird that every other function starts with appID and AppType as the first two parameters, and this one starts with the 'name' parameter.  Updating this for consistency would be a good idea, unless there's a strong reason otherwise.

client/collector.go
smithers1221: My thought was that to call this you actual need to have created a spec and have a specID...which would be impossible if it was a kots app.

smithers1221: *actually*

PR #78, opened 2019-12-03
add --ensure-channel flag for kots apps

dexhorthy:
Now you can do something like

```shell
replicated release create \
  --yaml-dir=./manifests/ \
  --promote=${GITHUB_BRANCH} \
  --ensure-channel
```

And it will promote the release to the `${GITHUB_BRANCH}` channel, creating that channel if it does not exist yet.

### DIFF
--- a/cli/cmd/release_create.go
+++ b/cli/cmd/release_create.go
@@ -36,17 +36,28 @@ func (r *runners) InitReleaseCreate(parent *cobra.Command) {
 	cmd.Flags().StringVar(&r.args.createReleasePromoteNotes, "release-notes", "", "When used with --promote <channel>, sets the **markdown** release notes")
 	cmd.Flags().BoolVar(&r.args.createReleasePromoteRequired, "required", false, "When used with --promote <channel>, marks this release as required during upgrades.")
 	cmd.Flags().StringVar(&r.args.createReleasePromoteVersion, "version", "", "When used with --promote <channel>, sets the version label for the release in this channel")
+	cmd.Flags().BoolVar(&r.args.createReleasePromoteEnsureChannel, "ensure-channel", false, "When used with --promote <channel>, will create the channel if it doesn't exist")
 
 	cmd.RunE = r.releaseCreate
 }
 
 func (r *runners) releaseCreate(cmd *cobra.Command, args []string) error {
 	if r.args.createReleaseYaml == "" && r.args.createReleaseYamlFile == "" && r.args.createReleaseYamlDir == "" {
-		return errors.New("yaml is required")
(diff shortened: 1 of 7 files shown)


pkg/kotsclient/channel.go
```diff
@@ -15,8 +15,14 @@ type GraphQLResponseGetChannel struct {
...
 }
+
+type GraphQLResponseCreateChannel struct {
+	Data   *KotsGetChannelData `json:"data,omitempty"`
+	Errors []graphql.GQLError  `json:"errors,omitempty"`
+}
+
 type KotsGetChannelData struct {
-	KotsChannel *KotsChannel `json:"getKotsChannel"`
+	KotsChannel *KotsChannel `json:"createKotsChannel"`
```
emosbaugh: Weird naming

cli/cmd/runner.go
```diff
@@ -41,27 +41,28 @@ type runnerArgs struct {
...
+	createReleaseDeploymentYaml       string
+	createReleaseServiceYaml          string
+	createReleasePreflightYaml        string
+	createReleaseSupportBundleYaml    string
+	createReleasePromote              string
+	createReleasePromoteDir           string
+	createReleasePromoteRequired      bool
+	createReleasePromoteNotes         string
+	createReleasePromoteVersion       string
+	createReleasePromoteEnsureChannel bool
```
dexhorthy: this is the new field

dexhorthy:
Tested this locally against production, and tested an AKA application as well.

pkg/kotsclient/channel.go
dexhorthy: oh i might have goofed this, good catch

cli/cmd/release_create.go
```diff
@@ -166,3 +163,42 @@ func (r *runners) releaseCreate(cmd *cobra.Command, args []string) error {
...
+		}
+	}
+
+	if len(promoteChannelIDs) == 0 {
+
+		// get-or-create only supported for KOTS apps at the moment
+		shouldCreateChannel := r.args.createReleasePromoteEnsureChannel && r.appType == "kots"
+
+		if shouldCreateChannel {
+			channelID, err := r.kotsAPI.CreateChannel(r.appID, r.args.createReleasePromote, "")
```
smithers1221: is this arg createReleasePromote or createReleasePromoteDir for Kots app?

dexhorthy: `createReleasePromote` is the *channel name* to promote the release to after creating -- I'd be open to updating the variable name in a followup PR if you think there's a way to make that more clear

dexhorthy:
@smithers1221 I like the idea of messaging that to users though. will add

PR #80, opened 2019-12-03
add a basic pact test for the kots client

laverya:
this requires acting as a new pact consumer, unfortunately

### DIFF
--- a/pkg/kotsclient/release_test.go
+++ b/pkg/kotsclient/release_test.go
@@ -1,0 +1,71 @@
+package kotsclient
+
+import (
+	"fmt"
+	"net/url"
+	"testing"
+
+	"github.com/pact-foundation/pact-go/dsl"
+	"github.com/replicatedhq/replicated/pkg/graphql"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_ListKotsReleasesActual(t *testing.T) {
+	var test = func() (err error) {
+		u := fmt.Sprintf("http://localhost:%d/graphql", pact.Server.Port)
+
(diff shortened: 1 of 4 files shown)


pkg/kotsclient/release_test.go
```diff
@@ -0,0 +1,71 @@
...
+		}).
+		WillRespondWith(dsl.Response{
+			Status: 200,
+			Body: map[string]interface{}{
+				"data": map[string]interface{}{
+					"allKotsReleases": dsl.EachLike(map[string]interface{}{
+						"sequence": dsl.Like(dsl.Integer()),
+						"created":  dsl.Like(dsl.String("Tue Nov 10 2009 23:00:00 UTC")),
+						"channels": dsl.EachLike(map[string]interface{}{
+							"id":             dsl.Like(dsl.String("id")),
```
salahalsaleh1993: would be good to check for the exact id value if possible and not just `dsl.Like`, cuz that will only check that it's a string

pkg/kotsclient/release_test.go
```diff
@@ -0,0 +1,71 @@
...
+					"appId": "kots-app",
+				},
+			},
+		}).
+		WillRespondWith(dsl.Response{
+			Status: 200,
+			Body: map[string]interface{}{
+				"data": map[string]interface{}{
+					"allKotsReleases": dsl.EachLike(map[string]interface{}{
+						"sequence": dsl.Like(dsl.Integer()),
```
salahalsaleh1993: same here, and for other important fields like ids, sequences, etc...

PR #94, opened 2020-01-21
another refactor, add --fail-on flag, fix double-printing errors

dexhorthy:
CC @austinchambers  @aj-jester  @Vwampage for visibility

```
$ go run ./cli release lint --yaml-dir=../kots-sentry/manifests
RULE                           TYPE    LINE    MESSAGE    
config-option-password-type    warn    65      Config option "external_postgres_secretname" should have type "password"
$ echo $?
0
```

Failure with warning

```
$ go run ./cli release lint --yaml-dir=../kots-sentry/manifests --fail-on=warn
RULE                           TYPE    LINE    MESSAGE    
config-option-password-type    warn    65      Config option "external_postgres_secretname" should have type "password"
Error: One or more errors of severity "warn" or higher were found
$ echo $?
1
```

Unsupported threshold:

```
$ go run ./cli release lint --yaml-dir=../kots-sentry/manifests --fail-on=lol
Error: fail-on value "lol" not supported, supported values are [info, warn, error, none]
```

Using `""` or `none` when `error` level messages are present:

```
$ go run ./cli release lint --yaml-dir=../kots-sentry/manifests --fail-on=""
RULE                TYPE     LINE    MESSAGE    
msg-yaml-invalid    error    2       end of the stream or a document separator is expected at line 2, column 1:
    }
    ^
config-spec                warn              Missing config spec
config-option-not-found    warn    16    Config option "hostname" not found
config-option-not-found    warn    20    Config option "postgres_type" not found
config-option-not-found    warn    24    Config option "external_postgres_host" not found
config-option-not-found    warn    25    Config option "external_postgres_password" not found
config-option-not-found    warn    26    Config option "external_postgres_port" not found
$ echo $?
0
```

PR #96, opened 2020-01-21
Release lint

dexhorthy:


dexhorthy:
This was approved here: https://github.com/replicatedhq/replicated/pull/94

but I made it against the wrong base

PR #102, opened 2020-03-06
Fix 'release update --yaml-dir'

### LINKED ISSUE #101: Replicated update fails with 'no such file or directory'
Trying to update an existing release with
```
replicated release update N --yaml-dir .
```
fails with
```
Error: read yaml dir: walk : lstat : no such file or directory
```

MikaelSmith:
Uses the correct variable name to get the `--yaml-dir` argument. Previously it would always try to walk an empty string path.

Fixes #101.

Signed-off-by: Michael Smith <michael.smith@puppet.com>

### DIFF
--- a/cli/cmd/release_update.go
+++ b/cli/cmd/release_update.go
@@ -59,7 +59,7 @@ func (r *runners) releaseUpdate(cmd *cobra.Command, args []string) error {
 	}
 
 	if r.args.updateReleaseYamlDir != "" {
-		r.args.updateReleaseYaml, err = readYAMLDir(r.args.createReleaseYamlDir)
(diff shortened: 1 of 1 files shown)


dexhorthy:
Nice catch @MikaelSmith -- thanks for the patch!

dexhorthy:
@MikaelSmith FYI right now the TravisCI only works on the upstream repo, but I've pushed this commit to github.com/replicatedhq/replicated/tree/fix-release-update and verified that it's passing there: https://travis-ci.org/replicatedhq/replicated/builds/659280355?utm_source=github_status&utm_medium=notification

Thanks again for the fix, we'll look to release 0.20.2 in the next few days!

PR #112, opened 2020-04-03
Assign and unassign channels

marccampbell:


cli/print/enterprise_channel.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/replicatedhq/replicated/pkg/enterprisetypes"
 )
 
-var enterpriseChannelTmplSrc = `ID	NAME	# VENDORS
+var enterpriseChannelTmplSrc = `ID	NAME	# VENDORS	# POLICIES
 {{ range . -}}
 {{ .ID }}	{{ .Name }}
```
divolgin, line 12: missing 1 item

marccampbell: :nod:

PR #117, opened 2020-04-09
Create and list enterprise installers routes

salahalsaleh:


pkg/enterprisetypes/installer.go
```diff
@@ -0,0 +1,6 @@
+package enterprisetypes
+
+type Installer struct {
+	ID   string
+	Yaml string
```
marccampbell, line 5: Does gofmt want to name this YAML instead?

salahalsaleh: nope

PR #119, opened 2020-04-13
Moving to go 1.14 and modules

marccampbell:
A few updates to the project:

1. use go 1.14.2 now
2. use gomodules for dependency management (set in Makefile)
3. make build should not copy the binary out of the path, this allows a local dev environment that doesn't overwrite a globally installed replicated binary


Dockerfile
```diff
@@ -1,15 +1,11 @@
...
 
-RUN go get golang.org/x/tools/cmd/goimports
-
-RUN go get github.com/spf13/cobra/cobra
-
 RUN go get github.com/go-swagger/go-swagger/cmd/swagger
 
 WORKDIR $PROJECTPATH
 
+ENV GO111MODULE=on
```
laverya, line 9: this is redundant with go 1.14

PR #134, opened 2020-05-12
include a timestamp and nonce in signed enterprise requests

laverya:
the old signature format is also still provided to prevent breakage, and an exported function has been created to validate signatures

### DIFF
--- a/pkg/enterpriseclient/auth.go
+++ b/pkg/enterpriseclient/auth.go
@@ -260,8 +260,8 @@ func combineTsNonceData(ts, nonce, data []byte) []byte {
 	return tsBytes
 }
 
-// ValidatePayload checks that the payload was signed by the provided public key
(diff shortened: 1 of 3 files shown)


pkg/enterpriseclient/auth.go
```diff
@@ -191,26 +194,124 @@ func homeDir() string {
...
+
+func combineTsNonceData(ts, nonce, data []byte) []byte {
+	tsBytes := []byte{}
+	tsBytes = append(tsBytes, nonce...)
+	tsBytes = append(tsBytes, ts...)
+	tsBytes = append(tsBytes, data...)
+	return tsBytes
+}
+
+// ValidatePayload checks that the payload was signed by the provided public key
```
salahalsaleh, line 263: signed by the public key? don't you mean the private key? or maybe i don't get the phrasing

laverya: yeah, bad phrasing
"signed by the private key associated with the provided public key" or something

PR #139, opened 2020-08-09
add command to fetch a license

dexhorthy:


pkg/platformclient/client.go
```diff
@@ -70,10 +71,15 @@ func (c *HTTPClient) doJSON(method, path string, successStatus int, reqBody, res
 		return fmt.Errorf("%s %s %d: %s", method, endpoint, resp.StatusCode, body)
 	}
 	if respBody != nil {
-		if err := json.NewDecoder(resp.Body).Decode(respBody); err != nil {
+		bodyBytes, err := ioutil.ReadAll(resp.Body)
```
dexhorthy, line 74: I know this is less efficient to read all the bytes in one go, but I had changed this so I could view the json body to debug deserialization issues, and I think that extra debug value might be worth the performance hit.

What do you think?

divolgin: it's never a lot of data

dexhorthy: my thoughts as well -- thanks

PR #141, opened 2020-08-10
restore method that got dropped due to merge conflict

dexhorthy:


pkg/platformclient/client.go
```diff
@@ -83,3 +83,29 @@ func (c *HTTPClient) DoJSON(method, path string, successStatus int, reqBody, res
 	return nil
 }
 
+
+func (c *HTTPClient) HTTPGet(path string, successStatus int) ([]byte, error) {
+
+	endpoint := fmt.Sprintf("%s%s", c.apiOrigin, path)
+	var buf bytes.Buffer
```
emosbaugh, line 90: cant you just pass in nil here?

dexhorthy: oh good catch

PR #142, opened 2020-08-10
fixing brew warning in goreleaser, auto-generating release properties, move build to GHA

dexhorthy:
Pulling some of the logic from the Makefile into the CLI here.

No special support for GitHub actions yet, but this shouldn't break any situation in which folks are using the Makefile and its special GHA logic to release.

```

$ go run ./cli release create --auto
  • Reading Environment ✓

Prepared to create release with defaults:

    yaml-dir        "./manifests"
    promote         "auto-release"
    version         "auto-release-180c2e-dirty"
    release-notes   "CLI release by dex on 11 Aug 20 15:44 CDT"
    ensure-channel  true

Create release with these properties? [Y/n] █

  • Reading manifests from ./manifests ✓
  • Creating Release ✓
    • SEQUENCE: 128
  • Promoting ✓
    • Channel 1fy7ktlXka3WylbLIR3w218CO1G successfully set to release 128

```


.github/workflows/release.yml
```diff
@@ -0,0 +1,19 @@
+name: Release CLI
+
+on:
+  create:
+    tags:
+      - 'v*'
+
+jobs:
```
dexhorthy, line 8: I haven't tested this but will try a release in a bit and see if it works, may have a followup PR

cli/cmd/release_create.go
```diff
@@ -4,13 +4,22 @@ import (
...
+	"github.com/replicatedhq/replicated/cli/print"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"gopkg.in/src-d/go-git.v4"
```
laverya, line 17: this should use `https://github.com/go-git/go-git/v5` instead - same API, new home, not an archived github repo

cli/print/logger.go
```diff
@@ -0,0 +1,233 @@
...
+)
+
+type Logger struct {
+	w io.Writer
+	spinnerStopCh chan bool
+	spinnerMsg    string
+	spinnerArgs   []interface{}
+	isSilent      bool
+	isVerbose     bool
+}
```
laverya, line 21: why write our own UI lib here?

dexhorthy: as discussed, poached from kots but needed to fork it so it could support a custom `io.Writer` https://github.com/replicatedhq/kots/blob/master/pkg/logger/logger.go

PR #149, opened 2020-08-26
add installer commands to kots channel inspect

dexhorthy:
this uses go template `with` to skip these on non-kots apps. I really want these install scripits to be sent down from the server side, but for now they are built by the client in vendor.replicated.com, so we'll build them in the client here as well.

```
$ replicated channel inspect dextestxxxxx
ID:             _-0LiRWGjBG2-VtJIr0F127QPiuR7T6M
NAME:           dextestxxxxx
DESCRIPTION:
RELEASE:        11
VERSION:        0.0.1
EXISTING:

    curl -fsSL https://kots.io/install | bash
    kubectl kots install kots-dex/dextestxxxxx

EMBEDDED:

    curl -fsSL https://k8s.kurl.sh/kots-dex-dextestxxxxx | sudo bash

AIRGAP:

   curl -fsSL -o kots-dex-dextestxxxxx.tar.gz https://k8s.kurl.sh/bundle/kots-dex-dextestxxxxx.tar.gz
   tar xvf kots-dex-dextestxxxxx.tar.gz
   sudo bash ./install.sh airgap

```

Other stuff
----------

- Move KOTS ListChannels call to vendor v3 API, this involved passing app slug down into a bunch of methods
- added appSlug to `*cmd.runner` during the Pre-Run function
- refactored the common "find channel by name" logic onto `Client` instead of duplicating it on each client impl

### DIFF
--- a/cli/print/channel_attributes.go
+++ b/cli/print/channel_attributes.go
@@ -1,22 +1,33 @@
 package print
 
 import (
+	"github.com/replicatedhq/replicated/pkg/types"
 	"text/tabwriter"
 	"text/template"
-
-	channels "github.com/replicatedhq/replicated/gen/go/v1"
 )
 
-var channelAttrsTmplSrc = `ID:	{{ .Id }}
+var channelAttrsTmplSrc = `ID:	{{ .ID }}
 NAME:	{{ .Name }}
 DESCRIPTION:	{{ .Description }}
 RELEASE:	{{ if ge .ReleaseSequence 1 }}{{ .ReleaseSequence }}{{else}}	{{end}}
-VERSION:	{{ .ReleaseLabel }}
+VERSION:	{{ .ReleaseLabel }}{{ with .InstallCommands }}
+EXISTING:
+
+{{ .Existing }}
+
+EMBEDDED:
+
+{{ .Embedded }}
+
+AIRGAP:
+
+{{ .Airgap }}
+{{end}}
 `
 
 var channelAttrsTmpl = template.Must(template.New("ChannelAttributes").Parse(channelAttrsTmplSrc))
 
-func ChannelAttrs(w *tabwriter.Writer, appChan *channels.AppChannel) error {
+func ChannelAttrs(w *tabwriter.Writer, appChan *types.Channel) error {
 	if err := channelAttrsTmpl.Execute(w, appChan); err != nil {
 		return err
 	}
--- a/client/channel.go
+++ b/client/channel.go
@@ -118,7 +118,7 @@ func (c *Client) findChannel(channels []types.Channel, name string) (*types.Chan
 	matchingChannels := make([]*types.Channel, 0)
 	for _, channel := range channels {
 		if channel.ID == name || channel.Name == name {
-			matchingChannels = append(matchingChannels, &channel)
+			matchingChannels = append(matchingChannels, channel.Copy())
 		}
 	}
 	if len(matchingChannels) == 0 {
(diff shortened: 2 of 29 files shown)


client/channel.go
```diff
@@ -1,13 +1,14 @@
...
-	"errors"
+	"fmt"
+	"github.com/pkg/errors"
 
 	channels "github.com/replicatedhq/replicated/gen/go/v1"
 	"github.com/replicatedhq/replicated/pkg/types"
 )
 
-func (c *Client) ListChannels(appID string, appType string) ([]types.Channel, error) {
+func (c *Client) ListChannels(appID string, appType string, appSlug string) ([]types.Channel, error) {
```
dexhorthy, line 11: you're gonna start seeing appSlug passed in everywhere because the vendor/v3 endpoint uses it to list channels.

client/channel.go
```diff
@@ -80,21 +81,52 @@ func (c *Client) CreateChannel(appID string, appType string, name string, descri
...
 		}
-		return c.KotsClient.ListChannels(appID)
+		return c.KotsHTTPClient.ListChannels(appID, appSlug)
 	}
 
 	return nil, errors.New("unknown app type")
 }
 
-func (c *Client) GetChannelByName(appID string, appType string, name string, description string, createIfAbsent bool) (*types.Channel, error) {
+func (c *Client) GetOrCreateChannelByName(appID string, appType string, appSlug string, name string, description string, createIfAbsent bool) (*types.Channel, error) {
```
dexhorthy, line 90: This method is refactored from each individual Platform/Ship/KOTS client, because the logic is basically the same.

pkg/kotsclient/channel.go
```diff
@@ -41,6 +43,58 @@ type KotsChannel struct {
...
+
+	slug := fmt.Sprintf("%s-%s", appSlug, c.ChannelSlug)
+	if c.ChannelSlug == "stable" {
+		slug = appSlug
+	}
+	kurlURL := fmt.Sprintf("%s/bundle/%s.tar.gz", kurlBaseURL, slug)
+
+	return fmt.Sprintf(`    curl -fsSL -o %s.tar.gz %s
+    tar xvf %s.tar.gz
+    sudo bash ./install.sh airgap`, slug, kurlURL, slug)
```
dexhorthy, line 84: I'm just spitballin over here folks but this seems useful

pkg/kotsclient/channel.go
```diff
@@ -95,6 +149,38 @@ query getKotsAppChannels($appId: ID!) {
 }
 `
 
+type ListChannelsResponse struct {
+	Channels []*KotsChannel `json:"channels"`
+}
+
+func (c *HTTPClient) ListChannels(appID string, appSlug string) ([]types.Channel, error) {
```
dexhorthy, line 156: here's the new vendor v3 impl, it also attaches the install commands 🙂

cli/print/channel_attributes.go
```diff
@@ -1,22 +1,33 @@
...
 import (
+	"github.com/replicatedhq/replicated/pkg/types"
 	"text/tabwriter"
 	"text/template"
-
-	channels "github.com/replicatedhq/replicated/gen/go/v1"
 )
 
-var channelAttrsTmplSrc = `ID:	{{ .Id }}
+var channelAttrsTmplSrc = `ID:	{{ .ID }}
```
dexhorthy, line 9: uhhhhh so apparently this breaks a bunch of tests...?

dexhorthy: ahh no i think we changed some channels.AppChannel to types.Channel, need to dig into that. stay tuned

go.mod
```diff
@@ -27,4 +27,5 @@ require (
 	golang.org/x/crypto v0.0.0-20200728195943-123391ffb6de
 	golang.org/x/sys v0.0.0-20200413165638-669c56c373c4 // indirect
 	gopkg.in/yaml.v2 v2.3.0 // indirect
+        github.com/ulikunitz/xz v0.5.8
```
laverya, line 30: this feels improperly formatted
not really a problem

dexhorthy: no it was bothering me too I'll fix it

PR #153, opened 2020-09-15
add --auto option for kurl installers, update release notes generation for release creation to match starter makefile

dexhorthy:
$ go run ./cli installer create --auto -y

          • Reading Environment ✓

        Prepared to create release with defaults:

            yaml-file       "./kurl-installer.yaml"
            promote         "release-and-installer"
            ensure-channel  true

          • Creating Release ✓
            • SEQUENCE: 32
          • Promoting ✓
            • Channel 1hY2rbsbOTIUkwt9jo297FDAG2y successfully set to release 32

And for updated release notes in release creation

    $ go run ./cli release create --auto -y
          • Reading Environment ✓

        Prepared to create release with defaults:

            yaml-dir        "./manifests"
            promote         "release-and-installer"
            version         "release-and-installer-b3a311-dirty"
            release-notes   "CLI release of release-and-installer triggered by dex [SHA: b3a311-dirty] [15 Sep 20 07:46 CDT]"
            ensure-channel  true

          • Reading manifests from ./manifests ✓
          • Creating Release ✓
            • SEQUENCE: 137
          • Promoting ✓
            • Channel 1hY2rbsbOTIUkwt9jo297FDAG2y successfully set to release 137

And GH actions:

    $ export GITHUB_ACTIONS=1; export GITHUB_REF=refs/heads/master; export GITHUB_SHA=1234567890; export GITHUB_REPOSITORY=replicatedhq/replicated-starter-kots; export GITHUB_ACTOR=dexhorthy;

    $ go run ./cli release create --auto  
    • Reading Environment ✓  

    Prepared to create release with defaults:

        yaml-dir        "./manifests"
        promote         "release-and-installer"
        version         "release-and-installer-c94b75-dirty"
        release-notes   "GitHub Action release of refs/heads/master triggered by dexhorthy: [1234567](https://github.com/replicatedhq/replicated-starter-kots/commit/1234567890)"
        ensure-channel  true


dexhorthy:
LGTM'd by @aj-jester  in slack (but @austinchambers I still wanna get your thoughts if you have any)

austinchambers:
Originally, I raised my eyebrows at the `--auto` flag. It seems less clear to me. There's the release that includes only manifest changes, and then there's the release that includes manifest changes _and_ installer changes. So this really felt like another parameter for the `release create` usage. 

Then I realized that I'm using `release` in the typical sense. With Replicated, there are five types of releases: 
- An "application" release object (application manifest changeset update, creating a new app release object) 
- An "installer" release object (installer manifest changeset update, creating a new installer release object)
- An "application" release (promoting an app release object)
- An "installer" release (promoting an installer release object)
- The "release" from the perspective of the Admin Console (which is a combination of all of the above at a given version). 

If we have so many different types of releases, then I suppose it's necessary to introduce a new CLI mode to handle the specifics separately. In this way, you have a compromise: 
- `release create` for app manifest changeset and promotion
- `installer create` for installer manifest changeset and promotion
- `--auto` for either of the above to tie them together as an actual packaged `release`

I wish we only had one type of release we express to the end user...   Since we can't have that, I get this approach. However, I think it might be a little more clear to have syntax like: 
`installer create --with-release`
and 
`release create --with-installer`

This way, we're a little more explicit (as `auto` can mean a lot of things).

dexhorthy:
@austinchambers -- I think I'm totally with you on "too many kinds of releases" -- Do you think it would be a step in the right direction to allow a "kurl installer" yaml for embedded cluster to bundled in with `manifests` like everything else?

For this, are you suggesting we eventually make `auto` default to true? That would be my goal

> As for auto specifically, we somewhat have the convention to use defaults when they are not specified (e.g., with our kots cli). Why can't we do that here?

austinchambers:
@dexhorthy Yeah, if there isn't a reason otherwise, we can document all the default values that auto sets in the CLI usage (like we do for kots CLI), and then simply omit the `-auto` flag entirely. I think that's better. 

And yes, I absolutely believe that it's better to imagine a kurl installer manifest as part of the application manifests. In fact, with my talks with Marc, I believe he's aligned on this. So maybe it makes sense to make the CLI act in that fashion, until vendorweb has a chance to catch up. With that in mind, that would argue for removing the `installer create` CLI commands entirely and instead make everything work within `release create`

PR #163, opened 2020-11-23
Return error when app slug is missing.

tdensmore:
Just a small change to cut my teeth on a simple change.

### DIFF
--- a/pkg/kotsclient/app.go
+++ b/pkg/kotsclient/app.go
@@ -105,5 +105,5 @@ func (c *GraphQLClient) GetApp(appID string) (*types.App, error) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/kotsclient/app.go
```diff
@@ -105,5 +105,5 @@ func (c *GraphQLClient) GetApp(appID string) (*types.App, error) {
 		}
 	}
 
-	return nil, errors.New("App not found")
+	return nil, errors.New("App not found" + appID)
```
dexhorthy, line 108: Do we need a space between "found" and appID

tdensmore:
This change is included in the channel name fix.

PR #165, opened 2020-11-30
platform channel search returns ErrNotFound if not found, not gql error string

laverya:
so we need to check for this too


divolgin (approved):

dexhorthy:
@laverya lgtm

divolgin (approved):

PR #166, opened 2020-12-11
Todd fix promote channel

tdensmore:
Small change so that the promote command:

`replicated release promote 14 todd --version 0.0.14`

will output human readable channel name:

`Channel todd successfully set to release 14`

instead of the current:

`Channel 1lTUjrHZchcezP6lUWhBwrdMcbg successfully set to release 14`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ logs/
 .DS_Store
 bin/
 vendor/
+.idea
--- a/pkg/kotsclient/app.go
+++ b/pkg/kotsclient/app.go
@@ -105,5 +105,5 @@ func (c *GraphQLClient) GetApp(appID string) (*types.App, error) {
 		}
 	}
 
-	return nil, errors.New("App not found")
+	return nil, errors.New("App not found" + appID)
 }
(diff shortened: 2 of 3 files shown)


tdensmore:
Fixed the bad error message.

tdensmore:
Let me know if this looks good to you @laverya

pkg/kotsclient/app.go
```diff
@@ -105,5 +105,5 @@ func (c *GraphQLClient) GetApp(appID string) (*types.App, error) {
 		}
 	}
 
-	return nil, errors.New("App not found")
```
laverya, line 108: I don't like this (string concat in general)

I'd prefer for this to be `fmt.Errorf("app %q not found", appID)` or similar instead, but it's not a blocker

.gitignore
```diff
@@ -5,3 +5,4 @@ logs/
 .DS_Store
 bin/
 vendor/
+.idea
```
laverya, line 8: I just add this to my global gitignore - see https://stackoverflow.com/a/7335487/5657035

laverya (approved):

PR #168, opened 2020-12-29
fix env var to match default in kots starter

dexhorthy:
https://github.com/replicatedhq/replicated-starter-kots/blob/main/.github/workflows/main.yml#L36


tdensmore (commented):
LGTM

PR #170, opened 2021-03-24
Javier add create lint auto

soriaj:
- Added ```--lint``` flag to ```replicated release create``` command 
- Adding ```--lint``` flag to ```--auto``` or ```--auto -y``` will set the lint-release value to true

### DIFF
--- a/cli/cmd/release_create.go
+++ b/cli/cmd/release_create.go
@@ -16,6 +16,8 @@ import (
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	// import go linter  
+	"github.com/replicatedhq/replicated/cli/cmd"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


cli/cmd/release_create.go
```diff
@@ -146,11 +151,25 @@ func (r *runners) setKOTSDefaultReleaseParams() error {
...
+		r.args.lintReleaseFailOn = "none"
+	} else if r.args.lintReleaseFailOn == "info" {
+		r.args.lintReleaseFailOn = "info"
+	} else if r.args.lintReleaseFailOn == "warn" {
+		r.args.lintReleaseFailOn = "warn"
+	} else if r.args.lintReleaseFailOn == "error" {
+		r.args.lintReleaseFailOn = "error"
+	} else {
+		return errors.New("You have entered an incorrect fail-on flag. See 'replicated release create --help' for valid options.")
+	}
```
dexhorthy, line 167: I thought the linter command we're delegating to handles this logic already? Have you tried passing an invalid option, without this logic included?

soriaj: The lint is set to false by default. If ```--lint``` is provided it will only fail on an 'error'.  If the user provides the ```--fail-on``` flag along with ```--lint``` it will check to ensure that they are valid options (i.e. to check for misspelled words).

dexhorthy: right, but doesn't the downstream command we call [here](https://github.com/replicatedhq/replicated/pull/170/files#diff-dc2fd422e3ea0093695faa8716b4fcb019cb1e2044ffc143cdb3b94380dd1995R219) have that same logic in it [over here](https://github.com/replicatedhq/replicated/blob/master/cli/cmd/release_lint.go#L48-L50)?

soriaj: You are correct. The release_lint.go contains the check for valid ```--fail-on``` options. Removing the ```--fail-on``` checks in the setKOTSDefaultReleaseParams will default fail-on to 'error' when the ```--lint``` flag is provided.

dexhorthy: okay cool -- so do we still need this logic in here or can it be removed?

cli/cmd/release_create.go
```diff
@@ -172,8 +177,9 @@ Prepared to create release with defaults:
     version         %q
     release-notes   %q
     ensure-channel  %t
+    lint-release    %t
```
dexhorthy, line 180: would it make sense to also show the `fail-on` value here?

soriaj: Possibly if lint-release is true. Otherwise fail-on is ignored.

dexhorthy: I like that idea

dexhorthy (approved):
new changes look good to me 🙂

cli/cmd/runner.go
```diff
@@ -60,6 +60,8 @@ type runnerArgs struct {
 	createReleasePromoteNotes         string
 	createReleasePromoteVersion       string
 	createReleasePromoteEnsureChannel bool
+	// Add Create Release Lint
+	createReleaseLint			      bool
```
laverya, line 64: this appears to not be formatted

laverya (approved):

PR #174, opened 2021-04-06
fix tests

dexhorthy:


laverya (approved):

cli/test/kots_app_test.go
```diff
@@ -93,34 +93,13 @@ var _ = Describe("kots apps", func() {
...
-			var stdout bytes.Buffer
-			var stderr bytes.Buffer
-
-			rootCmd := cmd.GetRootCmd()
-			rootCmd.SetArgs([]string{"app", "create", newName})
-
-			err = cmd.Execute(rootCmd, nil, &stdout, &stderr)
-			req.NoError(err)
-
-			req.Empty(stderr.String(), "Expected no stderr output")
```
laverya, line 113: IMO these need to get included after the 'app create' below

laverya (changes requested):
when combining the tests the new one should have the same checks as the old

cli/test/kots_app_test.go
dexhorthy: good call thank you

laverya (approved):
looks good to me - reconciles tests with new output format and combines two tests to cover the same code but be more consistent

PR #179, opened 2021-06-17
fix: return an error when a filename is used with the yaml flag

morningvera:
Partial fix for this [issue](https://app.clubhouse.io/replicated/story/34321/replicated-vendor-cli-cannot-promote-installer-releases-to-a-channel). Added an error case when users pass a yaml file as the argument to the `--yaml` flag. Also updated some of the error handling (using `pkg/errors` exclusively instead of mixing it with Golang errors).

### DIFF
--- a/cli/cmd/collector_create.go
+++ b/cli/cmd/collector_create.go
@@ -3,7 +3,9 @@ package cmd
 import (
 	"fmt"
 	"io/ioutil"
+	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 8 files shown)


cli/cmd/collector_create.go
```diff
@@ -16,25 +18,28 @@ func (r *runners) InitCollectorCreate(parent *cobra.Command) {
...
-		return fmt.Errorf("yaml is required")
+		return errors.New("one of --yaml or --yaml-file is required")
 	}
 
 	if r.args.createCollectorYaml != "" && r.args.createCollectorYamlFile != "" {
-		return fmt.Errorf("only one of yaml or yaml-file may be specified")
+		return errors.New("only one of yaml or yaml-file may be specified")
+	}
+
+	if strings.Contains(r.args.createCollectorYaml, ".yaml") {
```
divolgin: Any string can appear anywhere in the input though.  How about testing if input is a file name like
```
os.Stat(r.args.createCollectorYaml)
```

morningvera: That's definitely an improvement but it'll fail if the filename doesn't exist. So for example, it'll still return `Kubernetes version is required` if the file isn't found. Seems a little confusing to me but maybe that's not a big deal. For now, I modified it to check for the `.yaml` / `.yml` suffix and to check that the input is a single word. I can change it to the `os.Stat` method if you prefer it though.

divolgin (approved):

DanStough (approved):

PR #180, opened 2021-08-04
base64 decode helm charts when downloading a kots release

sgalsaleh:


divolgin (approved):
Would it be worth to check if matching HelmChart spec exists instead of extracting all archives?

divolgin (approved):

PR #181, opened 2021-08-19
Jeffgolden/ch35161/migrate graphql to vendor api

jeffreygolden:


jeffreygolden:
Requires https://github.com/replicatedhq/vandoor/pull/1465 for tests to pass

pkg/types/installer.go
```diff
@@ -13,3 +15,41 @@ type InstallerSpec struct {
...
+	Yaml string `json:"yaml"`
+}
+
+type PromoteInstallerRequest struct {
+	Sequence     int64  `json:"sequence"`
+	VersionLabel string `json:"versionLabel"`
+	ChannelID    string `json:"channelId"`
+}
+
+type EnterpriseChannel struct {
```
sgalsaleh, line 33: why are these here? graphql requests don't interact with the enterprise api

jeffreygolden: they were provisional during dev, forgot to remove them after I found they weren't needed.  Removed now.

PR #189, opened 2022-02-14
Allow Y and N

jdewinne:


dexhorthy (dismissed):
@jdewinne lgtm -- wanna add a `type::` label?

divolgin:
I think `strings.ToLower()` makes it cleaner and could also be used for "yes" and "no"

jdewinne:
@divolgin How would the `strings.ToLower()` work for the `yes` use case? Or you thinking of using a combination of the `strings.ToLower` and the `map`

divolgin:
> How would the strings.ToLower() work for the yes

Go already has built-in primitives to handle this; the map is a hack.  `strings.ToLower()`allows you to reduce the number of cases by normalizing input with.  So for example `"Yes"`, `"YES"`, `"yeS"`, etc all become `"yes"`.  Similar to how `"Y"` and `"y"` both become just one case `"y"`.

> Or you thinking of using a combination of the strings.ToLower and the map

I think map is a hack.  Go already has built-in primitives for dealing with multiple cases.

```
switch input {
case "y", "Y", "n", "N":
  return nil
default:
  return errors.New(`please choose "y" or "n"`)
}
```

cli/cmd/release_create.go
```diff
@@ -184,7 +184,7 @@ Prepared to create release with defaults:
 			if err != nil {
 				return err
 			}
-			if confirmed != "y" {
+			if strings.ToLower(confirmed) == "n" {
```
divolgin, line 187: this condition should not be reversed

jdewinne: Sure. But why? an `==` is for me more readable vs `!=`

dexhorthy: I think this was originally written so that the default would be to decline. E.g. "only `yes` will be accepted" and bail in any other case. Sort of a double-check against any bugs in the prompt validation logic.

dexhorthy (approved):

dexhorthy (approved):

PR #190, opened 2022-02-24
Kiraboyle/sc 42504/add or remove semantic versioning attribute

kcboyle:


pkg/kotsclient/channel.go
```diff
@@ -160,3 +160,23 @@ func (c *VendorV3Client) ArchiveChannel(appID, channelID string) error {
 
 	return nil
 }
+
+func (c *VendorV3Client) UpdateSemanticVersioning(appID string, channel *channels.AppChannel, enableSemver bool) error {
+	request := types.UpdateChannelRequest{
```
sgalsaleh, line 165: from the reference page, it seems like the app id, channel id and name are all required, but i only see the name passed here. https://replicated-vendor-api.readme.io/v3/reference/updatechannel

pkg/types/channel.go
```diff
@@ -59,6 +59,15 @@ type CreateChannelRequest struct {
 	Name                       string `json:"name"`
 }
 
+type UpdateChannelRequest struct {
+	// Description of the channel that is to be created.
+	Name                  string `json:"name"`
+	SemverRequired        bool   `json:"semverRequired,omitempty"`
+	Description           string `json:"description,omitempty"`
+	ChannelIcon           string `json:"channelIcon,omitempty"`
+	AutoAirgapBuildsValue int    `json:"autoAirgapBuildsValue,omitempty"`
```
sgalsaleh, line 68: IMO unused fields shouldn't added here

pkg/kotsclient/channel.go
kcboyle: ah. I got lazy and passed the whole channel instead. If it's more clear I'll pass them both separately

pkg/types/channel.go
kcboyle: done

sgalsaleh (dismissed):

sgalsaleh (approved):

PR #197, opened 2022-03-11
Update goreleaser

jdewinne:


tdensmore:
LGTM.

.goreleaser.yml
```diff
@@ -12,8 +12,9 @@ brews:
   folder: HomebrewFormula
   install: bin.install "replicated"
 universal_binaries:
-- replace: true
-  id: replicated
+- ids:
+  - cli
```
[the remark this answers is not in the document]
jdewinne: This will break the install. Putting `cli`, names the binary also `cli`, but `bin.install` points to `replicated`.
But if you put `replicated` under `ids`, it doesn't create a `darwin_all` artifact.

tdensmore:
I think this might be the golden PR. Any eyeballs to approve?

damuret (dismissed):

.goreleaser.yml
```diff
@@ -12,8 +12,9 @@ brews:
...
+  replace: true
```
jdewinne, line 17: I think it should be
```
universal_binaries:
  - replace: true
    id: cli 
    name_template: replicated
```
At least that looks correct local for me. @tdensmore Can you verify?

jdewinne (commented):
See comments

.goreleaser.yml
tdensmore: Yes ID is deprecated in favor of IDS so this looks correct

dexhorthy:
@jdewinne @tdensmore if y'all think this is good to go, let's give it a shot.

dexhorthy (dismissed):
I don't fully grok but let's give it a shot

jdewinne:
> I don't fully grok but let's give it a shot

I disagree. This is not going to work, because the binary is named `cli`.

tdensmore:
LGTM - I think we need @dexhorthy or @damuret to approve.

PR #216, opened 2022-11-16
Adding option to enable airgap, gitops or snapshots

jdewinne:


dexhorthy (approved):
No issues with this implementation. At a certain point this parameter list gets pretty long and it might make sense to use a struct instead of 7+ positional params

PR #219, opened 2022-12-14
fix release workflow

e3b0c442:
* Remove needs reference to job that no longer exists
* Move can-i-deploy to the correct place


jdewinne (approved):
Not sure what `can-i-deploy` does, but is probably at the correct spot

PR #224, opened 2023-01-22
small refactor to allow library usage to set the license type

dexhorthy:


jdewinne (approved):
At some point we should refactor the long list of arguments

dexhorthy:
@jdewinne yep, agree 100%. thanks for the review

PR #225, opened 2023-02-02
show version update

marccampbell:
This adds some more info to the version command:

## Version Command
When the version is the latest (no change):

```
% replicated version
replicated version v0.41.0
```

When the version is not the latest, but the user installed via a GitHub release, or some other manual method:

```
% replicated version
replicated version v0.41.0
Update available: v0.42.0
To automatically upgrade, run "replicated version upgrade"
```

When the version is not the latest but the user installed via brew:

```
% replicated version
replicated version v0.41.0
Update available: v0.42.0
To upgrade, run "brew upgrade replicatedhq/replicated/cli"
```

## All other commands
This also adds version/upgrade checking into all other commands

When the version is the latest (no change):

```
$ replicated app ls
ID                             NAME               SLUG               SCHEDULER
2A2MTUE9fj2n9gX2nT6o9yw6RLF    Slackernews        slackernews        kots
```

When the version is not the latest, but the user installed via a GitHub release, or some other manual method:
```
% replicated app ls
Update available: v0.41.0
To automatically upgrade, run "replicated version upgrade"
ID                             NAME               SLUG               SCHEDULER
2A2MTUE9fj2n9gX2nT6o9yw6RLF    Slackernews        slackernews        kots
```

When the version is not the latest but the user installed via brew:
```
% replicated app ls
Update available: v0.41.0
To upgrade, run "brew upgrade replicatedhq/replicated/cli"
ID                             NAME               SLUG               SCHEDULER
2A2MTUE9fj2n9gX2nT6o9yw6RLF    Slackernews        slackernews        kots
```


pkg/version/version.go
```diff
@@ -68,3 +82,27 @@ func getGoInfo() GoInfo {
 		Arch:     runtime.GOARCH,
 	}
 }
+
+func Print() {
+	fmt.Printf("replicated version %s\n", build.Version)
+
+	if build.UpdateInfo != nil {
+		fmt.Printf("Update available: %s\n", build.UpdateInfo.LatestVersion)
```
divolgin, line 90: In the examples this is also printed with structured output.  How about sending these to stderr instead?

divolgin (approved):

PR #234, opened 2023-03-30
cluster ls: fix output

CpuID:
crashes with nothing otherwise

prototype code, that isn't going to main right now. feature branch to a longer running dev branch


CpuID:
forgot this was uncommitted work locally when I merged https://github.com/replicatedhq/replicated/pull/233 :(

PR #242, opened 2023-04-19
Fix for service account token to read from file

jdewinne:


cli/cmd/registry_add_gcr_test.go
```diff
@@ -0,0 +1,73 @@
...
+		},
+		{
+			name:          "service account key required",
+			args:          []string{"registry", "add", "gcr", "--endpoint", "gcr.io"},
+			wantErr:       true,
+			wantErrString: "serviceaccountkey or serviceaccountkey-stdin must be specified",
+		},
+		{
+			name:          "invalid service account key",
+			args:          []string{"registry", "add", "gcr", "--endpoint", "gcr.io", "--serviceaccountkey", "./testdata/invalid-gcr-service-account-key.json"},
```
divolgin, line 30: This test should not use relative file paths.  We can use`embed.FS` from the `https://pkg.go.dev/embed` package

Code should be refactored to validate regardless of whether JSON is coming from a file or from STDIN.

beck-at-replicated:
For context:

This bug prevents vendors from connecting to Google Container Registry securely. Both `replicated registry add` and the GUI interface result in json parsing errors of the private key. Our [documentation](https://community.replicated.com/t/how-do-i-add-private-gcp-google-container-registry-gcr-images-in-my-application/125/2) indicates that the error is correctable with modifications to the whitespace of the key. 

See [this slack thread ](https://replicated.slack.com/archives/C043S005AUC/p1686151152250909)for an example.


image_720

PR #244, opened 2023-05-01
Adding json output to customer and cluster comamnds

jdewinne:


cli/print/customers.go
```diff
@@ -14,9 +15,23 @@ var customersTmplSrc = `ID	NAME	CHANNELS	EXPIRES	TYPE
...
+		return w.Flush()
+	}
+	if outputFormat == "json" {
+		defer w.Flush()
+		var cAsByte []byte
+		if len(customers) == 1 {
+			cAsByte, _ = json.MarshalIndent(customers[0], "", "  ")
+		} else {
+			cAsByte, _ = json.MarshalIndent(customers, "", "  ")
+		}
```
TyBrown, line 32: I'd recommend against this sort of choice - this means that the output type is different depending on if there's 1 customer (hashmap), or >1 customer (array)... which makes it harder to write a script or output parser that can handle both scenarios.

cli/print/clusters.go
```diff
@@ -15,19 +16,41 @@ var clustersTmplSrc = `ID	NAME	K8S DISTRO	K8S VERSION	STATUS	CREATED	EXPIRES
...
+
+	}
+	if outputFormat == "json" {
+		defer w.Flush()
+		var cAsByte []byte
+		if len(clusters) == 1 {
+			cAsByte, _ = json.MarshalIndent(clusters[0], "", "  ")
+		} else {
+			cAsByte, _ = json.MarshalIndent(clusters, "", "  ")
+		}
```
TyBrown, line 34: Same here - changing the outer type in the JSON makes composable scripts harder to write.

cli/print/customers.go
jdewinne: So always return it as an array?

TyBrown: That would be my preference, yes. :)

jdewinne: Nevermind. We discussed this. I'll find a better way.

TyBrown: Adding summary from Zoom call - this function is called for both `customer get $CUSTOMER` and `customer ls` type commands, which would be a desirable use case for having two separate output styles (it makes sense for a "get" type request to output the single customer as a hashmap, and the "ls" style request to output an array always).

jdewinne: Better: It is used for both `customer create` and `customer ls`. There is no `customer get` command at the moment.
I've split it into 2 different print functions.

marccampbell (approved):

TyBrown (dismissed):

bbawn (dismissed):

cli/print/clusters.go
```diff
@@ -15,19 +16,54 @@ var clustersTmplSrc = `ID	NAME	K8S DISTRO	K8S VERSION	STATUS	CREATED	EXPIRES
...
-	return w.Flush()
+func Cluster(outputFormat string, w *tabwriter.Writer, cluster *types.Cluster) error {
+	if outputFormat == "table" {
+		if err := clustersTmpl.Execute(w, []types.Cluster{*cluster}); err != nil {
+			return err
+		}
+		return w.Flush()
+
+	}
+	if outputFormat == "json" {
```
divolgin, line 61: Would be good to see some consistency in style.  Some of these functions use `if` and some `else if`.

cli/print/customers.go
```diff
@@ -14,9 +15,34 @@ var customersTmplSrc = `ID	NAME	CHANNELS	EXPIRES	TYPE
...
-func Customers(w *tabwriter.Writer, customers []types.Customer) error {
-	if err := customersTmpl.Execute(w, customers); err != nil {
+func Customers(outputFormat string, w *tabwriter.Writer, customers []types.Customer) error {
+	if outputFormat == "table" {
+		if err := customersTmpl.Execute(w, customers); err != nil {
+			return err
+		}
+		return w.Flush()
+	} else if outputFormat == "json" {
+		defer w.Flush()
```
divolgin, line 25: The caller should be calling `Flush`

divolgin (changes requested):
Can this be changed to use `Flush` in a consistent way?

divolgin (approved):

PR #251, opened 2023-05-23
Lists customers running a specific version of an application

St0rmz1:
The already API provides instance data including the application version. However, its cumbersome to find which instances or customers that are running a specific software release. This subcommand allows you to use the CLI to report on customers (and their instances), which are running a specific version.

**Use Case**
A vendor wants to know which of their customers and instances are running a specific version of their software.

**Implementation**
- Use the existing Vendor API /v3/app/%s/customers? route to retrieve all customers
- Iterate the response json, looking at which customers are running the appID specified in the 
- Search the Instance versionHistory.VersionLabel for matching instances of the specified version
- Print the results to screen either table or json format (depending on if --output was provided. defaults to table if not provided)
- Utilizes all the existing Cobra, Runner, kotsclient, etc

**Usage**
Added a new sub command under customer called `search-by-app-version`. For example, the command can be called like:
`replicated customer search-by-app-version <VERSION> --app <APPID> --output json`

**Summary of Changes**
- New CLI command customer_by_app_version
- Extended the current `Customer` type to now include `Instances` and `VersionHistory`
- Added `appVersion` string to runner
- New print function `CustomersWithInstances` which shows customer, instance ID, Last Active, Version

### DIFF
--- a/cli/cmd/customer_by_app_version.go
+++ b/cli/cmd/customer_by_app_version.go
@@ -1,13 +1,14 @@
 package cmd
 
 import (
+	"sort"
+
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/replicated/cli/print"
 	"github.com/replicatedhq/replicated/pkg/kotsclient"
 	"github.com/spf13/cobra"
 )
 
-// replicated customer search-by-app-version APPID VERSION
 func (r *runners) InitCustomersByAppVersionCommand(parent *cobra.Command) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:          "search-by-app-version VERSION --app APPID",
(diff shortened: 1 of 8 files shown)


cli/cmd/runner.go
```diff
@@ -181,5 +181,8 @@ type runnerArgs struct {
 
 	lsClusterHideTerminated bool
 
+	// Application Version String
+	appVersion string
```
marccampbell, line 185: convention on this project is to prefix these vars with the command that uses them

cli/cmd/customer_by_app_version.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/replicatedhq/replicated/cli/print"
+	"github.com/replicatedhq/replicated/pkg/kotsclient"
+	"github.com/spf13/cobra"
+)
+
+func (r *runners) InitCustomersByAppVersionCommand(parent *cobra.Command) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "search-by-app-version VERSION --app APPID",
```
marccampbell, line 14: is is possible to make this just a filter flag on customer ls?

For example:

`replicated customer ls --app-version 2.2.2 --app=<slug/id>`?

No new subcommand needed this way?

St0rmz1:
closing old draft that we opted not to implement

PR #255, opened 2023-05-26
Add an 'api get' command

marccampbell:
This is a quick implementation of a `replicated api <path>` command that works for GET methods today, using the CLI creds.

Exmaple

```
% make build && ./bin/replicated api get v3/apps | jq | more
go build \
                -ldflags " -X github.com/replicatedhq/replicated/pkg/version.version=v0.44.0-22-g65b01545 -X github.com/replicatedhq/replicated/pkg/version.gitSHA=`git rev-parse HEAD` -X github.com/replicatedhq/replicated/pkg/version.buildTime=`date -u +"%Y-%m-%dT%H:%M:%SZ"` " \
                -o bin/replicated \
                cli/main.go
Update available: v0.46.0
To automatically upgrade, run "replicated version upgrade"
{
  "apps": [
    {
      "id": "cut",
      "teamId": "cut",
      "name": "Slackernews",
      "created": "2022-06-02T20:53:54Z",
      "description": "",
      "isArchived": false,
      "slug": "slackernews",
      "renamedAt": null,
      "isKotsApp": true,
      "publicKeyPem": "-----BEGIN PUBLIC
  <cut>
```

### DIFF
--- a/cli/cmd/api.go
+++ b/cli/cmd/api.go
@@ -1,0 +1,17 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func (r *runners) InitAPICommand(parent *cobra.Command) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:    "api",
+		Short:  "Make ad-hoc API calls to the Replicated API",
+		Long:   ``,
+		Hidden: false,
+	}
+	parent.AddCommand(cmd)
+
+	return cmd
+}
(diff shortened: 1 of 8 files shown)


divolgin (approved):

z4ce:
These kind of utility functions are really nice for exploring the API without having to worry about the auth flow. For Replicated, right now, that auth flow is just a bearer token, but even to find that out requires looking it up. Doing `replicated api` requires just passing paths around and encourages people to go explore. Eventually if have a more sophisticated auth workflow that might be oauth based for users then it gets really handy vs having to try to do it yourself.

PR #257, opened 2023-06-01
Fixes #256 Channel Inspect default output format

### LINKED ISSUE #256: replicated channel inspect returns no output
From version `0.46` of the CLI, `replicated channel inspect <my-channel>` doesn't seem to return any output unless you specify a value for `--output`.

The help text says that `--output=text` is supposed to be the default, but you need to set it explicitly to get any output. Both `text` and `json` seem to work, it's when you don't specify it that you get no output.

jdewinne:
Uses different variable as others have default `table`

### DIFF
--- a/cli/cmd/runner.go
+++ b/cli/cmd/runner.go
@@ -183,4 +183,6 @@ type runnerArgs struct {
 
 	apiPostBody string
 	apiPutBody  string
+
+	channelInspectOutputFormat string
 }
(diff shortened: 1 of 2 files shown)


cli/cmd/runner.go
```diff
@@ -183,4 +183,6 @@ type runnerArgs struct {
 
 	apiPostBody string
 	apiPutBody  string
+
+	channelInspectOutputFormat string
```
divolgin, line 187: How does this work? I don't see this variable used anywhere.

jdewinne: It is used in the `channel inspect` command

divolgin (approved):

PR #259, opened 2023-06-06
Return an error if an invalid cluster was created

marccampbell:
Return an error if an invalid/supported distribution or version was requested for c11y

### DIFF
--- a/pkg/kotsclient/cluster_create.go
+++ b/pkg/kotsclient/cluster_create.go
@@ -2,6 +2,7 @@ package kotsclient
 
 import (
 	"net/http"
+	"strings"
 
 	"github.com/replicatedhq/replicated/pkg/types"
 )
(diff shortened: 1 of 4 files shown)


pkg/kotsclient/cluster_create.go
```diff
@@ -73,8 +81,19 @@ func (c *VendorV3Client) CreateCluster(opts CreateClusterOpts) (*types.Cluster,
...
+		if strings.Contains(err.Error(), " 400: ") {
+			// to avoid a lot of brittle string parsing, we make the request again with
+			// a dry-run flag and expect to get the same result back
+			ve := ValidationError{}
+			err := c.DoJSON("POST", "/v3/cluster?dry-run=true", http.StatusOK, reqBody, &ve)
+			if err != nil {
+				return nil, nil, err
+			}
+
+			return nil, &ve, nil
```
divolgin, line 93: lgtm, but this would be more Go-style error handling if this function returned a custom error type instead of the additional return value

```return nil, CreateClusterErrors{Errors: ve.Errors}```

divolgin (approved):

PR #260, opened 2023-06-08
Relax validation for Builders apps

divolgin:
Only validate that --lint flag is not used with Builders app releases. All other KOTS application inputs are allowed. Version label will be validated by the API.


dexhorthy:
Makes sense. Have you validated that if the API rejects the `--version` label, the error is shown helpfully to the user?

divolgin:
Yeah, all validation errors are returned with 400 status and are printed to the terminal.

bbawn (approved):

PR #261, opened 2023-06-09
Add newline to ouput

jdewinne:
[SC 77671](https://app.shortcut.com/replicated/story/77671/don-t-show-sign-at-the-end-of-the-replicated-cluster-ls-command-when-no-clusters-are-available)

### DIFF
--- a/cli/cmd/customer_download_license.go
+++ b/cli/cmd/customer_download_license.go
@@ -1,9 +1,11 @@
 package cmd
 
 import (
+	"fmt"
+	"io/ioutil"
+
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-	"io/ioutil"
 )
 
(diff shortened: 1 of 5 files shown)


bbawn (dismissed):

divolgin:
Can we change this to use `fmt.Fprintln`?

jdewinne:
> Can we change this to use `fmt.Fprintln`?

Maybe. I guess that should be a change for all commands, so we keep it consistent. Which might go beyond the scope of this PR?

divolgin:
I don't see anything else using `w.Write` and appending line breaks.  I do see `fmt.Fprintln` being used elsewhere.

jdewinne:
Done. And I've also added some others that were inconsistent.

divolgin (approved):

PR #268, opened 2023-06-21
Options to get kubeconfig

marccampbell:
Allows use to get kubeconfig by id or name of cluster (id as a position arg remains the default).  also allows writing the kubeconfig to a file instead of merging, if desired.

### DIFF
--- a/cli/cmd/cluster_kubeconfig.go
+++ b/cli/cmd/cluster_kubeconfig.go
@@ -27,9 +27,12 @@ func (r *runners) InitClusterKubeconfig(parent *cobra.Command) *cobra.Command {
 		Long:         `Download credentials for a test cluster`,
 		RunE:         r.kubeconfigCluster,
 		SilenceUsage: true,
-		Args:         cobra.MinimumNArgs(1),
 	}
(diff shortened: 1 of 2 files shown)


cli/cmd/cluster_kubeconfig.go
```diff
@@ -27,20 +27,62 @@ func (r *runners) InitClusterKubeconfig(parent *cobra.Command) *cobra.Command {
...
+	} else if r.args.kubeconfigClusterName != "" {
+		clusters, err := kotsRestClient.ListClusters(false)
+		if err != nil {
+			return errors.Wrap(err, "list clusters")
+		}
+		for _, cluster := range clusters {
+			if cluster.Name == r.args.kubeconfigClusterName {
+				clusterID = cluster.ID
+				break
+			}
```
jdewinne, line 58: There can be multiple clusters with the same name.
Ok to always take the first one. But I think in other cases (for example channel commands), we return an error that we can not uniquely define it.

jdewinne (approved):
Added a comment about names not being unique

PR #269, opened 2023-06-22
Memory should be using GiB

jdewinne:


bbawn (commented):
Implementation seems fine. I'm a little puzzled by the requirement though. I thought convention (k8es and elsewhere) would be `memoryBytes int64` and parse the suffix. A lot of k8es examples have memory < 1GiB. What's the rationale behind GiB granularity?

jdewinne:
> Implementation seems fine. I'm a little puzzled by the requirement though. I thought convention (k8es and elsewhere) would be `memoryBytes int64` and parse the suffix. A lot of k8es examples have memory < 1GiB. What's the rationale behind GiB granularity?

This is the memory for the MicroVM where the k8s cluster will run on. For example: if not specified by the end user the default is 4GB. I think you're thinking it is the memory for one of the pods? If so, that is not the case.

pkg/kotsclient/cluster_create.go
```diff
@@ -13,7 +13,7 @@ type CreateClusterRequest struct {
 	KubernetesVersion      string `json:"kubernetes_version"`
 	NodeCount              int    `json:"node_count"`
 	VCpus                  int64  `json:"vcpus"`
-	MemoryMiB              int64  `json:"memory_mib"`
+	MemoryMiB              int64  `json:"memory_gib"`
```
divolgin, line 16: `MemoryGiB`

divolgin (approved):

divolgin (approved):

PR #279, opened 2023-06-29
Adds functionality to display customers and their instances for a specific app version

St0rmz1:
The API already provides instance data including the application version. However, its cumbersome to find which instances or customers that are running a specific software release. This subcommand allows you to use the CLI to report on customers (and their instances), which are running a specific version.

**Use Case**
A vendor wants to know which of their customers and instances are running a specific version of their software.

**Implementation**
- Use the existing Vendor API /v3/app/%s/customers? route to retrieve all customers
- Iterate the response json, looking at which customers are running the appID specified in the 
- Search the Instance versionHistory.VersionLabel for matching instances of the specified version
- Print the results to screen either table or json format (depending on if --output was provided. defaults to table if not provided)

**Usage**
Added a new `--appversion` flag for the `customer ls` command to specify a version string

Example, 
`replicated customer ls  --app <APPID> --appversion <VERSION>`

**Summary of Changes**
modified:   cli/cmd/customer_ls.go
add --appversion flag
if appversion provided, then call ListCustomersByAppAndVersion()

modified:   cli/cmd/runner.go
add lsAppVersion variable

modified:   cli/print/customers.go
add CustomersWithInstances() which prints customer name and instances with the specific app version

modified:   client/customer.go
add ListCustomersByAppAndVersion()

new file:   pkg/kotsclient/customer_list_by_app_version.go

modified:   pkg/types/customer.go
add customer type to now include instance struct

new file:   pkg/types/instance.go
add new instance type

### DIFF
--- a/client/customer.go
+++ b/client/customer.go
@@ -21,7 +21,7 @@ func (c *Client) ListCustomers(appID string, appType string) ([]types.Customer, 
 func (c *Client) ListCustomersByAppAndVersion(appID string, appVersion string, appType string) ([]types.Customer, error) {
 
 	if appType == "platform" {
-		return nil, errors.New("listing customers is not supported for platform applications")
+		return nil, errors.New("listing customers by app version is not supported for native scheduler applications")
 	} else if appType == "kots" {
 		return c.KotsClient.ListCustomersByAppAndVersion(appID, appVersion, appType)
 	}
--- a/client/customer.go
+++ b/client/customer.go
@@ -17,6 +17,18 @@ func (c *Client) ListCustomers(appID string, appType string) ([]types.Customer, 
 	return nil, errors.Errorf("unknown app type %q", appType)
 }
 
+// list customers by app and app version
+func (c *Client) ListCustomersByAppAndVersion(appID string, appVersion string, appType string) ([]types.Customer, error) {
+
+	if appType == "platform" {
+		return nil, errors.New("listing customers is not supported for platform applications")
+	} else if appType == "kots" {
+		return c.KotsClient.ListCustomersByAppAndVersion(appID, appVersion, appType)
+	}
+
+	return nil, errors.Errorf("unknown app type %q", appType)
+}
+
 func (c *Client) CreateCustomer(appType string, opts kotsclient.CreateCustomerOpts) (*types.Customer, error) {
 	if appType == "platform" {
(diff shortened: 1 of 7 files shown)


client/customer.go
```diff
@@ -17,6 +17,18 @@ func (c *Client) ListCustomers(appID string, appType string) ([]types.Customer,
 	return nil, errors.Errorf("unknown app type %q", appType)
 }
 
+// list customers by app and app version
+func (c *Client) ListCustomersByAppAndVersion(appID string, appVersion string, appType string) ([]types.Customer, error) {
+
+	if appType == "platform" {
+		return nil, errors.New("listing customers is not supported for platform applications")
```
marccampbell, line 24: `listing customers by app version is not supported for native scheduler applications`

pkg/kotsclient/customer_list_by_app_version.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+		if len(matchingCustomers) == resp.TotalCustomers || len(resp.Customers) == 0 {
+			// We've reached the end of the customer pages
+			break
+		}
+		page = page + 1
+	}
+
+	if len(matchingCustomers) == 0 {
+		return matchingCustomers, fmt.Errorf("no matching customers found for app %q and version %q ", appID, appVersion)
```
marccampbell, line 53: if someone has `--output json`, will this throw an error instead of the expected return of `[]` ?

client/customer.go
St0rmz1: Changed text per suggestion. "listing customers by app version is not supported for native scheduler applications"

pkg/kotsclient/customer_list_by_app_version.go
St0rmz1: I addressed this in cli/cmd/customer_ls.go
When customers is 0 and output is json, we pass a blank struct to the print function, which will result in []
Otherwise, the output format is table, in which case, we raise an error instead.

// if err and outputFormat is json, customers should be a blank struct and print will return []
		if err != nil && r.outputFormat == "json" {
			return print.CustomersWithInstances(r.outputFormat, r.w, customers)

St0rmz1:
resolved merge conflicts

marccampbell (approved):

PR #289, opened 2023-07-19
Add CLI customer inspect command with helm instructions

bbawn:
https://app.shortcut.com/replicated/story/81196/cli-only-way-to-get-customer-helm-install-instructions

### DIFF
--- a/cli/cmd/customer_inspect.go
+++ b/cli/cmd/customer_inspect.go
@@ -1,0 +1,125 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/replicatedhq/replicated/cli/print"
+	"github.com/replicatedhq/replicated/pkg/types"
+)
+
+func (r *runners) InitCustomersInspectCommand(parent *cobra.Command) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:          "inspect",
+		Short:        "Show full details for a customer",
+		Long:         `Show full details for a customer`,
+		RunE:         r.inspectCustomer,
+		SilenceUsage: true,
+	}
+	parent.AddCommand(cmd)
+	cmd.Flags().StringVar(&r.args.customerInspectCustomer, "customer", "", "The Customer Name or ID")
+	cmd.Flags().StringVar(&r.args.customerInspectOutputFormat, "output", "text", "The output format to use. One of: json|text (default: text)")
+
+	return cmd
+}
+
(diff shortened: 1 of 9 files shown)


cli/cmd/customer_inspect.go
```diff
@@ -0,0 +1,125 @@
...
+	return nil
+}
+
+func (r *runners) customerChannel(customer *types.Customer) (*types.KotsChannel, error) {
+	var ch *types.Channel
+	if len(customer.Channels) == 0 {
+		return nil, nil
+	}
+	if len(customer.Channels) > 1 {
+		fmt.Println("WARNING: customer has multiple channels, using first channel")
```
divolgin, line 62: Since this command produces structured output, this should go to `fmt.Fprintf(os.Stderr, ...`

cli/cmd/customer_inspect.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// TODO: fix GET /v3/app/{appID}/customer/{customerID} and/or GET /v3/app/{appID}/channel/{channelID} to
+	// include chartNames in channels, like List does. Also ?channelName selector wrongly excludes details
+	channels, err := r.kotsAPI.ListKotsChannels(r.appID, "", false)
+	if err != nil {
+		return nil, errors.Wrapf(err, "list channels for customer %q", customer.Name)
+	}
+
+	for _, detailedChan := range channels {
+		if ch.ID == detailedChan.Id {
```
divolgin, line 74: How about a nested loop here to iterate customer channels instead of doing `customer.Channels[0]` with a warning message.

cli/print/customer_attributes.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	if len(preflightReleases) == 0 {
+		return "No preflight checks found"
+	}
+
+	var cmd string
+	for i, chart := range preflightReleases {
+		var delim string
+		if i < len(preflightReleases)-1 {
```
divolgin, line 99: these loops can be simplified with `strings.Join(cmds, " &&\n")`

cli/cmd/customer_inspect.go
bbawn: Yeah I was wondering about that. Is there currently a way to make a Customer with multiple Channels? Looks like v3 APIs for customer create/update just have a single ChannelID .

divolgin: We only allow one channel right now.  Multi-channel support has been on the back burner forever now.

divolgin (approved):

PR #291, opened 2023-07-26
cache update checker info

pavansokkenagaraj:
feat: 
- cache update checker info and build version 
- set cache ttl to 4 hours
- invalidate cache when
    * current version is not cached version
    * ttl expires 

chore: 
- update gitignore
- cleanup cluster prepare cmd helper
- switch to os package read/write

slack: https://replicated.slack.com/archives/C04T9AAMJUS/p1690383500550359?thread_ts=1690381446.326349&cid=C04T9AAMJUS

### DIFF
--- a/pkg/replicatedfile/cache.go
+++ b/pkg/replicatedfile/cache.go
@@ -30,7 +30,7 @@ func init() {
 	var err error
 	Cache, err = loadCache()
 	if err != nil {
-		fmt.Println(err)
+		fmt.Printf("Failed to load cache file: %v\n", err)
 	}
 }
 
--- a/pkg/replicatedfile/cache.go
+++ b/pkg/replicatedfile/cache.go
@@ -18,6 +18,7 @@ const (
 )
 
 type cache struct {
+	CachedVersion     string                    `json:"cachedVersion,omitempty"`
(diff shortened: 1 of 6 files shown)


cli/cmd/version.go
```diff
@@ -16,6 +16,10 @@ func Version() *cobra.Command {
 		Use:   "version",
 		Short: "Print the current version and exit",
 		Long:  `Print the current version and exit`,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			version.InitBuild()
```
emosbaugh, line 20: could we instead move NewUsrbinSDK into this function and leave init() the way it was so the version package is as expected?

emosbaugh (changes requested):

pkg/replicatedfile/cache.go
```diff
@@ -0,0 +1,119 @@
...
+
+var (
+	Cache *cache
+)
+
+func init() {
+	var err error
+	Cache, err = loadCache()
+	if err != nil {
+		fmt.Println(err)
```
emosbaugh, line 33: can you add some contectual information in this error? something like

```suggestion
		fmt.Printf("Failed to load cache file: %v\n", err)
```

emosbaugh (changes requested):

emosbaugh (approved):

PR #297, opened 2023-08-03
fix: remove appID required for cluster cmds

pavansokkenagaraj:
fix: 
- remove appID required for cluster cmds
- set appID from env `REPlCATED_APP` for `replicated cluster prepare` cmd


marccampbell:
Is there any way to use the prerunCmd only for cluster prepare, to make sure that this code exists in one place?

marccampbell (approved):

PR #298, opened 2023-08-03
remove vcpus and memory flags

marccampbell:
This starts to use the new r1.<size> instance types instead.

Before this PR (and other supporting PRs), VM based instances required a vcpus and memory flag while cloud instances required an instance-type flag. This was a bad experience and worse since we only supported powers of 2 for vcpus and balanced (1 by 4) instance specs. Now we have standard instance types and a consistent API and CLI to request VM instances.


divolgin (dismissed):

emosbaugh (changes requested):
you may want to update the copy here

```
	cmd.Flags().StringVar(&r.args.createClusterInstanceType, "instance-type", "", "the type of instance to use for cloud-based clusters (e.g. x5.xlarge)")
```

emosbaugh (approved):

PR #303, opened 2023-08-07
fix: support for non-semver versions in cluster create command

emosbaugh:


marccampbell (dismissed):

cli/cmd/cluster_create.go
```diff
@@ -82,20 +80,18 @@ func createCluster(kotsRestClient *kotsclient.VendorV3Client, opts kotsclient.Cr
 
 	if ve != nil && len(ve.Errors) > 0 {
 		if len(ve.SupportedDistributions) > 0 {
-			return nil, fmt.Errorf("%s\n\nSupported Kubernetes distributions and versions are:\n%s", errors.New(strings.Join(ve.Errors, ",")), supportedDistributions(ve.SupportedDistributions))
-		} else {
-			return nil, fmt.Errorf("%s", errors.New(strings.Join(ve.Errors, ",")))
+			print.ClusterVersions("table", r.werr, ve.SupportedDistributions)
```
divolgin, line 83: Why is this printing to stderr?

cli/cmd/runner.go
```diff
@@ -31,6 +31,7 @@ type runners struct {
 	dir              string
 	outputFormat     string
 	w                *tabwriter.Writer
+	werr             *tabwriter.Writer
```
divolgin, line 34: What is the use for this?

cli/cmd/cluster_create.go
jdewinne: Because the original implementation was also sending it to stderr.
@marccampbell ok if we change this to stdout for showing the supported ones. And only do the actual error to stderr?

cli/cmd/runner.go
jdewinne: Related to your other comment. If we don't send the supported distro's on validation error to stderr, we will not need this one.

cli/cmd/cluster_create.go
divolgin: I don't see where this line prints an error.  None of this reads like an error, except the output name

```
print.ClusterVersions("table", r.werr, ve.SupportedDistributions)
```

divolgin: That looks like supported distros should to go to STDOUT and the error returned below should go to STDERR.

jdewinne: Fixed

cli/cmd/runner.go
jdewinne: Removed

divolgin (approved):

PR #304, opened 2023-08-07
Remove all clusters

marccampbell:
This adds a new `--all` flag to cluster rm. It also requires a `--force` flag to remove a running cluster.

### DIFF
--- a/cli/cmd/cluster_rm.go
+++ b/cli/cmd/cluster_rm.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/replicated/pkg/platformclient"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


cli/cmd/cluster_rm.go
```diff
@@ -1,30 +1,57 @@
...
+
+		for _, cluster := range clusters {
+			err := r.kotsAPI.RemoveCluster(cluster.ID)
+			if errors.Cause(err) == platformclient.ErrForbidden {
+				return ErrCompatibilityMatrixTermsNotAccepted
+			} else if err != nil {
+				return errors.Wrap(err, "remove cluster")
+			} else {
+				fmt.Printf("removed cluster %s\n", cluster.ID)
+			}
```
jdewinne, line 49: Suggestion: extract this into its own function, so it can be the same and reused for L54-59

jdewinne (approved):

PR #310, opened 2023-08-11
feat: return human readable err when requesting kubeconfig for cluster not running 

pavansokkenagaraj:
return human readable err when requesting kubeconfig for cluster not running

https://app.shortcut.com/replicated/story/82999/requesting-kubeconfig-for-cluster-not-running-should-provide-a-nicer-human-readable-error

image

### DIFF
--- a/cli/cmd/cluster_kubeconfig.go
+++ b/cli/cmd/cluster_kubeconfig.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/replicatedhq/replicated/pkg/platformclient"
+	"github.com/replicatedhq/replicated/pkg/types"
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/tools/clientcmd"
 )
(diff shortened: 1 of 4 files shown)


cli/cmd/cluster_kubeconfig.go
```diff
@@ -64,6 +65,17 @@ func (r *runners) kubeconfigCluster(_ *cobra.Command, args []string) error {
 		return errors.New("must provide cluster id or name")
 	}
 
+	cluster, err := r.kotsAPI.GetCluster(clusterID)
+	if errors.Cause(err) == platformclient.ErrForbidden {
```
divolgin, line 69: I don't think that endpoint ever returns StatusForbidden

divolgin (approved):

cli/cmd/cluster_kubeconfig.go
pavansokkenagaraj: I will clean up in the next PR

PR #319, opened 2023-08-31
feat (c11y): cluster upgrades

emosbaugh:
```
$ replicated cluster upgrade --help
Upgrade a test clusters


Usage:
  replicated cluster upgrade [cluster id] [flags]

Flags:
      --dry-run          Dry run
  -h, --help             help for upgrade
      --output string    The output format to use. One of: json|table (default: table) (default "table")
      --version string   Kubernetes version to upgrade to (format is distribution dependent)
      --wait duration    Wait duration for cluster to be ready (leave empty to not wait)

Global Flags:
      --app string     The app slug or app id to use in all calls
      --token string   The API token to use to access your app in the Vendor API
```

### DIFF
--- a/cli/cmd/cluster_create.go
+++ b/cli/cmd/cluster_create.go
@@ -29,15 +29,15 @@ func (r *runners) InitClusterCreate(parent *cobra.Command) *cobra.Command {
 	cmd.Flags().IntVar(&r.args.createClusterNodeCount, "nodes", int(1), "Node count")
 	cmd.Flags().Int64Var(&r.args.createClusterDiskGiB, "disk", int64(50), "Disk Size (GiB) to request per node")
 	cmd.Flags().StringVar(&r.args.createClusterTTL, "ttl", "", "Cluster TTL (duration, max 48h)")
-	cmd.Flags().BoolVar(&r.args.createClusterDryRun, "dry-run", false, "Dry run")
 	cmd.Flags().DurationVar(&r.args.createClusterWaitDuration, "wait", time.Second*0, "Wait duration for cluster to be ready (leave empty to not wait)")
-	cmd.Flags().StringVar(&r.args.createClusterInstanceType, "instance-type", "", "The type of instance to use (e.g. x5.xlarge)")
-	cmd.Flags()
+
+	cmd.Flags().StringVar(&r.args.createClusterInstanceType, "instance-type", "", "the type of instance to use (e.g. x5.xlarge)")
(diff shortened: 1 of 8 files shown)


divolgin (approved):

cli/cmd/cluster_create.go
```diff
@@ -29,15 +29,15 @@ func (r *runners) InitClusterCreate(parent *cobra.Command) *cobra.Command {
...
+
+	cmd.Flags().StringVar(&r.args.createClusterInstanceType, "instance-type", "", "the type of instance to use (e.g. x5.xlarge)")
+	cmd.Flags().BoolVar(&r.args.createClusterDryRun, "dry-run", false, "Dry run")
 
 	cmd.Flags().StringVar(&r.outputFormat, "output", "table", "The output format to use. One of: json|table (default: table)")
 
-	cmd.MarkFlagRequired("distribution")
-	cmd.MarkFlagRequired("version")
+	_ = cmd.MarkFlagRequired("distribution")
+	_ = cmd.MarkFlagRequired("version")
```
jdewinne, line 40: Is adding `_` something we are planning on doing consistently?

pkg/kotsclient/cluster_upgrade.go
```diff
@@ -0,0 +1,65 @@
...
+		if strings.Contains(err.Error(), " 400: ") {
+			// to avoid a lot of brittle string parsing, we make the request again with
+			// a dry-run flag and expect to get the same result back
+			ve, _ := c.doUpgradeClusterDryRunRequest(clusterID, req)
+			if ve != nil && len(ve.Errors) > 0 {
+				return nil, ve, nil
+			}
+		}
+		return nil, nil, err
+	}
```
jdewinne, line 52: There is a potential better way, without having to assume `400:` in the error message.
https://github.com/replicatedhq/replicated/blob/main/pkg/kotsclient/release_compatibility.go#L40-L49
The `err` also contains the body response. So if it is a bad request, based on the swagger api, you could parse the returned body.

jdewinne (approved):
Added some comments. But LGTM

pkg/kotsclient/cluster_upgrade.go
emosbaugh: this is copy paste from cluster_create

cli/cmd/cluster_create.go
emosbaugh: i'd prefer we panic. ill propose that and add in a follow up

PR #335, opened 2023-10-23
Use builders linter if release has charts only

divolgin:
Implements a TODO comment.

If a release only has Helm charts, it should use Builders linter.

### DIFF
--- a/cli/cmd/release_lint.go
+++ b/cli/cmd/release_lint.go
@@ -1,7 +1,7 @@
 package cmd
 
 import (
-	"io/ioutil"
+	"io/fs"
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


cli/cmd/release_lint.go
```diff
@@ -127,10 +131,40 @@ func tarYAMLDir(yamlDir string) ([]byte, error) {
...
+		if err != nil {
+			return helmError
+		}
+
+		return nil
+	})
+
+	if err == helmError {
+		return false, nil
+	}
```
FourSigma, line 163: Do we want to return a `nil` error here?

divolgin: it's not an error.  it's the `false` result.

FourSigma:
```
_, err = loader.LoadFile(path)
```
We are basically using this call to see if this an Helm chart right? Is it possible returns other errors like "file not found"?  Additionally, do we need to wrap this helm error?

divolgin: If it can't be loaded then it's not a chart. There are no restrictions on what kinds of files can be passed in.

FourSigma (approved):

PR #338, opened 2023-11-02
Works around CLI expecting a 'paid' while API expects 'prod'

crdant:


cli/cmd/customer_create.go
```diff
@@ -47,6 +47,9 @@ func (r *runners) createCustomer(_ *cobra.Command, _ []string) (err error) {
 	if r.args.customerCreateType == "test" && r.args.customerCreateExpiryDuration > time.Hour*48 {
 		return errors.New("test licenses cannot be created with an expiration date greater than 48 hours")
 	}
+	if r.args.customerCreateType == "paid" {
```
dexhorthy, line 50: I'm not clear on what this fixes - what is stopping your from just sending `--prod` as the license type? Is this a documentation problem? Or just for parity with what the form in vendor portal says?

Screenshot 2023-11-03 at 9 02 23 AM

crdant: The fix could go either way…accept `prod` or convert to `paid`. But I think the CLI should match the GUI, and the GUI calls it "Paid".

crdant: I also think "Paid" is more of how the vendor will be thinking of it. "This is one of my paid customers" not "this is one of my prod customers".

divolgin (approved):

PR #340, opened 2023-11-08
CMX: Add additional node groups

marccampbell:


xavpaice:
I think we're safe to close this one now.

PR #353, opened 2024-01-17
Return timeout exit code in case of wait duration exceeded

jdewinne:
Will return exit code 124 in case the `--wait` flag is used and the cluster did not reach running status:

```
> replicated cluster create --distribution eks --wait 1m
ID          NAME                           DISTRIBUTION    VERSION       STATUS          CREATED                          EXPIRES    TAGS
f2317daa    bold_chatterjee                eks             1.28          provisioning    2024-01-17 00:56:52 +0000 UTC    -
> echo $?
124
```

### DIFF
--- a/cli/cmd/cluster_create.go
+++ b/cli/cmd/cluster_create.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/moby/moby/pkg/namesgenerator"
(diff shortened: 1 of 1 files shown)


xavpaice (approved):

cli/cmd/cluster_create.go
```diff
@@ -68,7 +71,13 @@ func (r *runners) createCluster(_ *cobra.Command, args []string) error {
 	}
 	cl, err := r.createAndWaitForCluster(opts)
 	if err != nil {
-		return err
+		if errors.Cause(err) == ErrWaitDurationExceeded {
+			defer func() {
+				os.Exit(124)
```
emosbaugh, line 76: wont this print "Dry run succeeded." since you defer here?

jdewinne: In theory, yes. But combining DryRun with Wait doesn't make a lot of sense. It actually should be impossible probably.

emosbaugh (approved):

PR #360, opened 2024-02-12
allow creating nodegroups on cluster create

jdewinne:


cli/cmd/cluster_create.go
```diff
@@ -143,3 +153,41 @@ func waitForCluster(kotsRestClient *kotsclient.VendorV3Client, id string, durati
 		time.Sleep(time.Second * 5)
 	}
 }
+
+func parseNodeGroups(nodeGroups []string) ([]kotsclient.NodeGroup, error) {
```
emosbaugh, line 157: i'd like to see a test for this function. also i imagine some fields are required here. should they be validated?

emosbaugh (changes requested):

emosbaugh (approved):

PR #368, opened 2024-03-05
Adding support for node autoscaling

jdewinne:


cli/cmd/cluster_create.go
```diff
@@ -35,11 +35,13 @@ https://docs.replicated.com/vendor/testing-how-to#limitations`,
 	cmd.Flags().StringVar(&r.args.createClusterKubernetesDistribution, "distribution", "", "Kubernetes distribution of the cluster to provision")
 	cmd.Flags().StringVar(&r.args.createClusterKubernetesVersion, "version", "", "Kubernetes version to provision (format is distribution dependent)")
 	cmd.Flags().IntVar(&r.args.createClusterNodeCount, "nodes", int(1), "Node count")
+	cmd.Flags().IntVar(&r.args.createClusterMinNodeCount, "min-nodes", int(-1), "Minimum Node count (only for EKS, AKS and GKE clusters)")
+	cmd.Flags().IntVar(&r.args.createClusterMaxNodeCount, "max-nodes", int(-1), "Maximum Node count (only for EKS, AKS and GKE clusters)")
```
emosbaugh, line 39: I think you should explain the value of -1 in the description

pkg/kotsclient/cluster_create.go
```diff
@@ -79,6 +85,13 @@ func (c *VendorV3Client) CreateCluster(opts CreateClusterOpts) (*types.Cluster,
 		Tags:                   opts.Tags,
 	}
 
+	if opts.MinNodeCount != -1 {
```
emosbaugh, line 88: What if it equals -2? Should this instead be >= 0?

emosbaugh (changes requested):

cli/cmd/cluster_create.go
```diff
@@ -35,8 +35,8 @@ https://docs.replicated.com/vendor/testing-how-to#limitations`,
 	cmd.Flags().StringVar(&r.args.createClusterKubernetesDistribution, "distribution", "", "Kubernetes distribution of the cluster to provision")
 	cmd.Flags().StringVar(&r.args.createClusterKubernetesVersion, "version", "", "Kubernetes version to provision (format is distribution dependent)")
 	cmd.Flags().IntVar(&r.args.createClusterNodeCount, "nodes", int(1), "Node count")
-	cmd.Flags().IntVar(&r.args.createClusterMinNodeCount, "min-nodes", int(-1), "Minimum Node count (only for EKS, AKS and GKE clusters)")
-	cmd.Flags().IntVar(&r.args.createClusterMaxNodeCount, "max-nodes", int(-1), "Maximum Node count (only for EKS, AKS and GKE clusters)")
+	cmd.Flags().IntVar(&r.args.createClusterMinNodeCount, "min-nodes", int(-1), "Minimum Node count (only for EKS, AKS and GKE clusters). A negative value will be ignored.")
```
emosbaugh, line 38: I particularly like how docker cli handles this in the run command

```
      --memory-swappiness int          Tune container memory swappiness (0 to 100) (default -1)
```

https://github.com/docker/cli/blob/a1361a1372adc0dff6016927a047bff71717a8d4/cli/command/container/opts.go#L304

emosbaugh (approved):
im approving this but I thought id share an example of what i thought was an elegant solution in a mature cli. feel free to disregard the suggestion if you do not agree

pkg/kotsclient/cluster_create.go
```diff
@@ -79,6 +85,13 @@ func (c *VendorV3Client) CreateCluster(opts CreateClusterOpts) (*types.Cluster,
 		Tags:                   opts.Tags,
 	}
 
+	if opts.MinNodeCount >= 0 {
```
emosbaugh, line 88: Sorry I think i steered you in the wrong direction here. This will unexpectedly drop a value if its invalid. I actually think you need to do some validation in the runners.createCluster function.

ideally what id like to see in that function is 

```
if r.args.createClusterMinNodeCount >= 0 {
  opts.MinNodeCount = r.args.createClusterMinNodeCount
} else if r.args.createClusterMinNodeCount != -1 {
  return errors.New("invalid value for flag --min-node-count (valid values are ...)", )
}
```

pkg/kotsclient/cluster_create.go
```diff
@@ -45,6 +49,8 @@ type NodeGroup struct {
 	Name         string `json:"name"`
 	InstanceType string `json:"instance_type"`
 	Nodes        int    `json:"node_count"`
+	MinNodes     int    `json:"min_node_count"`
+	MaxNodes     int    `json:"max_node_count"`
```
emosbaugh, line 53: I believe these need to be pointers to ints otherwise they will be 0 if unset

emosbaugh (changes requested):

emosbaugh (dismissed):

cli/cmd/cluster_create.go
```diff
@@ -79,6 +81,16 @@ func (r *runners) createCluster(_ *cobra.Command, args []string) error {
 		Tags:                   tags,
 		DryRun:                 r.args.createClusterDryRun,
 	}
+	if r.args.createClusterMinNodeCount >= 0 {
+		opts.MinNodeCount = &r.args.createClusterMinNodeCount
+	} else if r.args.createClusterMinNodeCount != -1 {
```
divolgin, line 86: Can this be a string input and the value parsed in code here?  Then just check if it's empty sting to use default values.

divolgin (approved):

PR #387, opened 2024-05-16
allow name flag for updating ttl

jdewinne:


marccampbell (approved):
we should note in the docs that this will match the _first_ cluster with a name

PR #400, opened 2024-07-07
Add total cost on cluster create dry run

jdewinne:


cli/cmd/cluster_create.go
```diff
@@ -114,7 +114,16 @@ func (r *runners) createCluster(_ *cobra.Command, args []string) error {
 	}
 
 	if opts.DryRun {
-		_, err := fmt.Fprintln(r.w, "Dry run succeeded.")
+		ttl := opts.TTL
+		if opts.TTL == "" {
+			ttl = "1h"
```
emosbaugh, line 119: This should be a flag default at this point to keep it consistent for both

marccampbell (dismissed):

pkg/types/cluster.go
```diff
@@ -28,6 +28,14 @@ type Cluster struct {
 	CreatedAt time.Time     `json:"created_at"`
 	ExpiresAt time.Time     `json:"expires_at"`
 
+	TTL string `json:"ttl"`
+
+	CreditsPerHourPerCluster int64 `json:"credits_per_hour_per_cluster"`
+	FlatFee                  int64 `json:"flat_fee"`
+	TotalCredits             int64 `json:"total_credits"`
+
+	EstimatedCost int64 `json:"_"`
```
emosbaugh, line 37: is this supposed to be a hyphen?

emosbaugh (changes requested):

pkg/types/cluster.go
```diff
@@ -28,6 +28,14 @@ type Cluster struct {
 	CreatedAt time.Time     `json:"created_at"`
 	ExpiresAt time.Time     `json:"expires_at"`
 
+	TTL string `json:"ttl"`
+
+	CreditsPerHourPerCluster int64 `json:"credits_per_hour_per_cluster"`
+	FlatFee                  int64 `json:"flat_fee"`
+	TotalCredits             int64 `json:"total_credits"`
+
+	EstimatedCost int64
```
emosbaugh, line 37: Not sure what you are going for here. Shouldn't this have some sort of json tag? If you were to list clusters --output json this probable doesn't look right the way it is.

emosbaugh (changes requested):

pkg/types/cluster.go
jdewinne: You're right. I was thinking as `EstimatedCost` is calculated (part of dry-run, or calculated for cluster list), it didn't have to go into json, but totally forgot about `--output`

emosbaugh (approved):

PR #401, opened 2024-07-18
Support for pushing and pulling ai models to and from the model repository

marccampbell:


cli/cmd/model_pull.go
```diff
@@ -0,0 +1,151 @@
...
+	// get the endpoint
+	endpoint, err := r.kotsAPI.GetModelsEndpoint()
+	if err != nil {
+		return err
+	}
+
+	if endpoint == "" {
+		return errors.New("Could not find models endpoint")
+	}
+	fmt.Printf("Pulling model from %s\n", endpoint)
```
divolgin, line 46: Is there a progress display when oras.Copy is called?

pkg/ociclient/upload.go
```diff
@@ -0,0 +1,217 @@
...
+	"net/http"
+	"os"
+	"time"
+
+	digest "github.com/opencontainers/go-digest"
+	specs "github.com/opencontainers/image-spec/specs-go"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+const chunkSize = 100 * 1024 * 1024 // 100 MB chunks
```
divolgin, line 20: How about changing this to upload the whole file until it chunking works in the registry correctly?

divolgin (approved):

cli/cmd/model_pull.go
marccampbell: not at this time

pkg/ociclient/upload.go
marccampbell: agreed, this only works for 100 MB files.  but larger gets really slow also.  
i'd prefer to keep this chunked and not working for > 100 MB than unknown size limitations.
product is not yet in prod

PR #404, opened 2024-07-30
Properly send content-range header and support multi-part uploads

marccampbell:
Send content range headers for multi-part upload

### DIFF
--- a/pkg/ociclient/upload.go
+++ b/pkg/ociclient/upload.go
@@ -17,7 +17,7 @@ import (
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
-const chunkSize = 100 * 1024 * 1024 // 100 MB chunks
+const chunkSize = 10 * 1024 * 1024 // 10 MB chunks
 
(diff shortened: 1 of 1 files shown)


pkg/ociclient/upload.go
```diff
@@ -17,7 +17,7 @@ import (
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
-const chunkSize = 100 * 1024 * 1024 // 100 MB chunks
+const chunkSize = 10 * 1024 * 1024 // 10 MB chunks
```
divolgin, line 20: this is just for testing, right?  10MB seems tiny.

divolgin (approved):

pkg/ociclient/upload.go
marccampbell: yeah, i think this needs to be a little larger (maybe 50 MB), but also more dynamic or user settable.  larger sizes are great for CI, but smaller will be good for poor client connections , right ?

PR #411, opened 2024-09-17
Use PATCH method to update channels

divolgin:
Use patch method for updating channels

### DIFF
--- a/pkg/types/channel.go
+++ b/pkg/types/channel.go
@@ -86,10 +86,8 @@ type CreateChannelRequest struct {
 	Name                       string `json:"name"`
 }
 
-type UpdateChannelRequest struct {
-	// Description of the channel that is to be created.
-	Name           string `json:"name"`
-	SemverRequired bool   `json:"semverRequired,omitempty"`
+type PatchChannelRequest struct {
+	SemverRequired *bool `json:"semverRequired,omitempty"`
 }
 
 type Channel struct {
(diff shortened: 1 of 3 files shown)


pkg/types/channel.go
```diff
@@ -86,10 +86,8 @@ type CreateChannelRequest struct {
 	Name                       string `json:"name"`
 }
 
-type UpdateChannelRequest struct {
-	// Description of the channel that is to be created.
-	Name           string `json:"name"`
-	SemverRequired bool   `json:"semverRequired,omitempty"`
+type PatchChannelRequest struct {
+	SemverRequired *bool `json:"semverRequired,omitempty"`
```
pandemicsyn, line 90: Doesn't this still need `name` ? In https://github.com/replicatedhq/vandoor/pull/6109/files#diff-92c0bd0fa66652f13bca297d2ede1533669dbf88e204083b2cba311270dd2f6d name is required?

```
	// Channel update parameters.
	// In: body
	Body struct {
		// Name of the channel that is to be updated.
		// Required: true
		Name *string `json:"name"`
```

pandemicsyn (changes requested):

pkg/types/channel.go
divolgin: No, it's optional. I'll update that swagger comment

pandemicsyn (approved):

PR #414, opened 2024-09-29
Joshd/sc 113095/api cli changes for vms

jdewinne:
Adding
```
replicated vm create
replicated vm ls
replicated vm versions
replicated vm rm
replicated vm update ttl
replicated vm group ls
```

### DIFF
--- a/cli/cmd/vm_nodegroup.go
+++ b/cli/cmd/vm_nodegroup.go
@@ -1,0 +1,14 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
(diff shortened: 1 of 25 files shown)


cli/cmd/vm_nodegroup.go
```diff
@@ -0,0 +1,14 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func (r *runners) InitVMNodeGroup(parent *cobra.Command) *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "nodegroup",
```
marccampbell, line 9: `nodegroup` is a very k8s term.  what is a better term here?

marccampbell (approved):

PR #418, opened 2024-10-09
Add update checking, performance improvements

marccampbell:
Bring the version checking and update inline, remove the cache implementation.


pkg/version/updateinfo.go
```diff
@@ -0,0 +1,119 @@
...
+	"encoding/json"
+	"net/http"
+	"time"
+
+	"github.com/Masterminds/semver"
+	"github.com/pkg/errors"
+)
+
+const (
+	latestVersionURI = `https://market-api-marccampbell.okteto.repldev.com/ping`
```
divolgin, line 13: hmm

pkg/version/updateinfo.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	pingResponse := PingResponse{}
+	err = json.NewDecoder(resp.Body).Decode(&pingResponse)
+	if err != nil {
+		return nil, errors.Wrap(err, "decode response")
+	}
+
+	return &VersionInfo{
+		Version: pingResponse.ClientVersions.ReplicatedCLI,
```
divolgin, line 77: Maybe make `ClientVersions` not a pointer and check CLI value:

```
if pingResponse.ClientVersions.ReplicatedCLI == "" {
  return nil, nil
}
```

divolgin (approved):

PR #426, opened 2024-10-10
Cache and default app

marccampbell:
Performance improvements:
- When we currently init the CLI on most commands, we hit the API to find the app. This introduces a cache where that API slug/id are stored.  there's no expiration on it right now, but these are immutable values generally?

Usability:
- Add a new "default" command:
`replicated default set app slackernews`
`replicated default show app`
`replicated default clear app`

These set the defaultApp value in the cache, so that we no longer need to pass it into every CLI function.

### DIFF
--- a/cli/cmd/default_show.go
+++ b/cli/cmd/default_show.go
@@ -1,0 +1,48 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/replicatedhq/replicated/cli/print"
+	"github.com/replicatedhq/replicated/pkg/types"
+	"github.com/spf13/cobra"
+)
+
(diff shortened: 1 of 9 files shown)


cli/cmd/default_show.go
```diff
@@ -0,0 +1,48 @@
...
+	cmd.Flags().StringVar(&outputFormat, "output", "table", "The output format to use. One of: json|table (default: table)")
+	parent.AddCommand(cmd)
+
+	return cmd
+}
+
+func (r *runners) showDefault(cmd *cobra.Command, defaultType string, outputFormat string) error {
+	if cache.DefaultApp == "" {
+		if outputFormat == "json" {
+			fmt.Println("{}")
```
divolgin, line 31: how about `null`?

divolgin (approved):

PR #427, opened 2024-10-10
bubble up more detailed git errors

pandemicsyn:
Right now when go-git encounters an error we end up not printing a very useful error since it ends being unwrapped. e.g.

```
ERROR: reference not found
```

With this change we should bubble up a more detailed error that should more directly hint at the problem:

```
ERROR: git resolve revision '"HEAD"' failed: reference not found
```

### DIFF
--- a/cli/cmd/release_create.go
+++ b/cli/cmd/release_create.go
@@ -109,7 +109,7 @@ func (r *runners) setKOTSDefaultReleaseParams() error {
 
 	rev, branch, isDirty, err := r.gitSHABranch()
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "get git properties")
 	}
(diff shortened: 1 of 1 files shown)


divolgin (approved):

cli/cmd/release_create.go
```diff
@@ -109,7 +109,7 @@ func (r *runners) setKOTSDefaultReleaseParams() error {
 
 	rev, branch, isDirty, err := r.gitSHABranch()
 	if err != nil {
-		return errors.Wrapf(err, "get git properties")
```
divolgin, line 112: i think you can keep this wrap, but `git` prefixes in the other errors make them trackable enough

divolgin (approved):

PR #450, opened 2024-10-18
Fix for git tree dirty

marccampbell:
Rename some variables.
Stop reusing the same error message.
And return true when it's ok.

### DIFF
--- a/dagger/release.go
+++ b/dagger/release.go
@@ -37,7 +37,7 @@ func (r *Replicated) Release(
 		return err
 	}
 	if !gitTreeOK {
-		return fmt.Errorf("Your git tree is not clean. You cannot release what's not committed.")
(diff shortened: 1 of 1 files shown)


dagger/release.go
```diff
@@ -246,15 +246,16 @@ func checkGitTree(ctx context.Context, source *dagger.Directory, githubToken *da
...
 		return false, err
 	}
 
-	output = strings.TrimSpace(output)
+	gitStatusOutput = strings.TrimSpace(gitStatusOutput)
 
-	if len(output) > 0 {
-		return false, ErrGitTreeNotClean
+	if len(gitStatusOutput) > 0 {
+		fmt.Printf("output: %s\n", gitStatusOutput)
```
pandemicsyn, line 257: 👍

pandemicsyn (approved):

PR #451, opened 2024-10-21
return ssh port and endpoint for direct connection

jdewinne:


marccampbell:
How can this work since we will not be sharing the public/private key with the customer?

jdewinne:
> How can this work since we will not be sharing the public/private key with the customer?

Because we add the github user's public key onto each vm at the moment:
https://github.com/replicatedhq/reliability-matrix/blob/main/cluster-provisioner/pkg/almalinux/cloudinit.go#L52

pkg/types/vm.go
```diff
@@ -20,8 +20,8 @@ type VM struct {
 
 	EstimatedCost int64 `json:"estimated_cost"` // Represents estimated credits for this vm based on the TTL
 
-	OverlayEndpoint string `json:"overlay_endpoint,omitempty"`
-	OverlayToken    string `json:"overlay_token,omitempty"`
+	SSHPort     int64  `json:"ssh_port,omitempty"`
```
marccampbell, line 23: i understand now.  we should maybe rename this to "direct ssh port" or something, since we will have a more supported ssh endpoint via CMX forwarder

marccampbell (approved):

PR #487, opened 2025-01-02
Support creating addons as part of 'cluster create' command

squizzi:
This adds a new flag to the `cluster create` command called `--addon`.  The addon flag is a `StringArrayVar` so it can be specified multiple times to add multiple types of addons, since we only support `object-store` at this time an example was added for that:

```
  # Create a cluster with addons
  replicated cluster create --distribution eks --version 1.21 --nodes 3 --addon object-store --ttl 24h
```

Since addons share the arguments for `ClusterID`, `OutputFormat`, `Duration` and `DryRun` those all get inherited by values used in the `clusterCreate` command.

The only unique flag is `bucket-prefix` which has been added as an argument to `clusterCreate` with some validation added to ensure it is used when the `addons` array contains `object-store`.

I also refactored the existing args into runnerArgs so the args can more easily shared across different commands.

# Testing
## Validation
```
bin/replicated cluster create --distribution eks --version 1.31 --nodes 3 --addon not-object-store --ttl 24h
Error: validate addon args: unknown addon: not-object-store

bin/replicated cluster create --distribution eks --version 1.31 --nodes 3 --addon object-store --ttl 24h
<Omitted cmd.Help printing>
Error: validate addon args: bucket-prefix is required for object-store addon
```

## Cluster Create
Works still without `--addon` specified:
```
bin/replicated cluster create --distribution eks --version 1.31 --nodes 3
ID          NAME                           DISTRIBUTION    VERSION       STATUS          CREATED                           EXPIRES                           COST
52db2783    silly_beaver                   eks             1.31          queued          2025-01-02 12:10 PST              -                                 $0.85
```

Adds an `object-store` when `--addon` specified:

```
bin/replicated cluster create --distribution eks --version 1.31 --nodes 3 --addon object-store --ttl 24h --bucket-prefix "banana"
ID          TYPE            STATUS          DATA
9b0ffa24    Object Store    pending         {"bucket_prefix":"banana"}
ID          NAME                           DISTRIBUTION    VERSION       STATUS          CREATED                           EXPIRES                           COST
6a7652cc    nostalgic_wing                 eks             1.31          queued          2025-01-02 12:13 PST              -                                 $8.79

# Addon goes ready
bin/replicated cluster addon ls 6a7652cc
ID          TYPE            STATUS          DATA
9b0ffa24    Object Store    ready           {"bucket_prefix":"banana","bucket_name":"banana-9b0ffa24-cmx","service_account_namespace":"cmx","service_account_name":"banana-9b0ffa24-cmx","service_account_name_read_only":"banana-9b0ffa24-cmx-ro"}
```

### DIFF
--- a/cli/cmd/cluster_addon_create_objectstore.go
+++ b/cli/cmd/cluster_addon_create_objectstore.go
@@ -13,17 +13,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type clusterAddonCreateObjectStoreArgs struct {
-	objectStoreBucket string
-	clusterID         string
-	waitDuration      time.Duration
-	dryRun            bool
-	outputFormat      string
-}
-
 func (r *runners) InitClusterAddonCreateObjectStore(parent *cobra.Command) *cobra.Command {
-	args := clusterAddonCreateObjectStoreArgs{}
-
 	cmd := &cobra.Command{
 		Use:   "object-store CLUSTER_ID --bucket-prefix BUCKET_PREFIX",
 		Short: "Create an object store bucket for a cluster.",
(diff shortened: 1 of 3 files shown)


squizzi:
> Error: validate addon args: unknown addon: not-object-store

It'd be nice to print a list of valid addons here, but for now we'd only be printing one.  We could do it by hand, ideally it'd be an API endpoint like we do for instance types that returns this info, but that seems overkill for right now.

jdewinne (approved):

PR #499, opened 2025-01-22
Add network create cli support

jdewinne:
Create a new network:
```
❯ replicated network create --ttl 1h
ID          NAME                           STATUS          CREATED                           EXPIRES
dbb9101a    unruffled_johnson              queued          2025-01-22 10:07 PST              -
```

List all networks:
```
❯ replicated network ls
ID          NAME                           STATUS          CREATED                           EXPIRES
dbb9101a    unruffled_johnson              running         2025-01-22 10:07 PST              2025-01-22 11:08 PST
8c9792ac    optimistic_shockley            running         2025-01-22 09:57 PST              2025-01-22 11:04 PST
```

Create a new vm, reusing an existing network:
```
❯ replicated vm create --distribution ubuntu --network dbb9101a
ID          NAME                           DISTRIBUTION    VERSION       STATUS          CREATED                           EXPIRES                           COST
f7cfc0ba    crazy_cray                     ubuntu          24.04         queued          2025-01-22 10:09 PST              -                                 $0.60
```

Associated Shortcut: [sc-116975](https://app.shortcut.com/replicated/story/116975/tailscale-set-each-tailnet-with-an-exit-node-and-make-tailscale0-be-the-default-route-interface)

### DIFF
--- a/cli/cmd/network_create.go
+++ b/cli/cmd/network_create.go
@@ -49,9 +49,7 @@ func (r *runners) createNetwork(_ *cobra.Command, args []string) error {
 	network, err := r.createAndWaitForNetwork(opts)
 	if err != nil {
 		if errors.Cause(err) == ErrVMWaitDurationExceeded {
-			defer func() {
-				os.Exit(124)
-			}()
+			defer os.Exit(124)
 		} else {
 			return err
 		}
@@ -100,13 +98,13 @@ func waitForNetwork(kotsRestClient *kotsclient.VendorV3Client, network *types.Ne
 
 		if network.Status == types.NetworkStatusRunning {
 			return network, nil
-		} else if network.Status == types.NetworkStatusError {
+		}
+		if network.Status == types.NetworkStatusError {
 			return nil, errors.New("network failed to provision")
-		} else {
-			if time.Now().After(start.Add(duration)) {
-				// In case of timeout, return the network and a WaitDurationExceeded error
-				return network, ErrWaitDurationExceeded
-			}
+		}
+		if time.Now().After(start.Add(duration)) {
+			// In case of timeout, return the network and a WaitDurationExceeded error
+			return network, ErrWaitDurationExceeded
 		}
 
 		time.Sleep(time.Second * 5)
--- a/cli/cmd/network_create.go
+++ b/cli/cmd/network_create.go
@@ -1,6 +1,16 @@
 package cmd
 
 import (
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/moby/moby/pkg/namesgenerator"
+	"github.com/pkg/errors"
+	"github.com/replicatedhq/replicated/cli/print"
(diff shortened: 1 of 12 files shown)


cli/cmd/network_create.go
```diff
@@ -14,9 +24,91 @@ func (r *runners) InitNetworkCreate(parent *cobra.Command) *cobra.Command {
...
+	opts := kotsclient.CreateNetworkOpts{
+		Name:   r.args.createNetworkName,
+		TTL:    r.args.createNetworkTTL,
+		DryRun: r.args.createNetworkDryRun,
+	}
+
+	network, err := r.createAndWaitForNetwork(opts)
+	if err != nil {
+		if errors.Cause(err) == ErrVMWaitDurationExceeded {
+			defer func() {
```
squizzi, line 52: Why is a `defer` needed?

Also, `os.Exit` doesn't return anything, so you don't need to wrap it in `func()` you can just do a one liner:

```
defer os.Exit(124)
```

cli/cmd/network_create.go
```diff
@@ -14,9 +24,91 @@ func (r *runners) InitNetworkCreate(parent *cobra.Command) *cobra.Command {
...
+				os.Exit(124)
+			}()
+		} else {
+			return err
+		}
+	}
+
+	if opts.DryRun {
+		_, err = fmt.Fprintln(r.w, "Dry run succeeded.")
```
squizzi, line 61: Can we call `print.Network` here as well so that we print the would be created network so the dry run is more then just a success/fail message?

```suggestion
		_, err = fmt.Fprintln(r.w, "Dry run succeeded. Network to be created:")
		print.Network(r.outputFormat, r.w, network)
```

cli/cmd/network_create.go
```diff
@@ -14,9 +24,91 @@ func (r *runners) InitNetworkCreate(parent *cobra.Command) *cobra.Command {
...
+
+func waitForNetwork(kotsRestClient *kotsclient.VendorV3Client, network *types.Network, duration time.Duration) (*types.Network, error) {
+	start := time.Now()
+	for {
+		network, err := kotsRestClient.GetNetwork(network.ID)
+		if err != nil {
+			return nil, errors.Wrap(err, "get network")
+		}
+
+		if network.Status == types.NetworkStatusRunning {
```
squizzi, line 101: :golf:  these are all returning something so wrapping them up in `else/if` shouldn't be necessary:

```go
		if network.Status == types.NetworkStatusRunning {
			return network, nil
		}

		if network.Status == types.NetworkStatusError {
			return nil, errors.New("network failed to provision")
		}

		if time.Now().After(start.Add(duration)) {
			// In case of timeout, return the network and a WaitDurationExceeded
			// error
			return network, ErrWaitDurationExceeded
		}
```

squizzi (approved):

divolgin (approved):

PR #512, opened 2025-02-27
Add ssh and scp subcommand

hedge-sparrow:
Currently SSH access to CMX VMs happens one of two ways:
- via the proxy
  - In some scenarios this can cause high latency at the terminal
- directly
  - but the user must fish out the ip and port number manually

This `replicated vm ssh` command handles connecting directly to a CMX vm via it's exposed port. 
it's behavior is such that: 
- specifying a VM by it's ID with `replicated vm ssh <id>` will connect you to that VM
- if no VMs are specified but multiple active VMs exist, prompt the user for which to connect to. 

TODO: 
- [ ] filter out airgapped VMs
- [x] add `scp` subcommand
- [ ] tests?

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -252,7 +252,6 @@ func Execute(rootCmd *cobra.Command, stdin io.Reader, stdout io.Writer, stderr i
 	runCmds.InitVMVersions(vmCmd)
 	runCmds.InitVMRemove(vmCmd)
 	runCmds.InitVMSSH(vmCmd)
-	runCmds.InitVMSCP(vmCmd)
 	vmUpdateCmd := runCmds.InitVMUpdateCommand(vmCmd)
 	runCmds.InitVMUpdateTTL(vmUpdateCmd)
 
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -252,6 +252,7 @@ func Execute(rootCmd *cobra.Command, stdin io.Reader, stdout io.Writer, stderr i
 	runCmds.InitVMVersions(vmCmd)
 	runCmds.InitVMRemove(vmCmd)
 	runCmds.InitVMSSH(vmCmd)
+	runCmds.InitVMSCP(vmCmd)
 	vmUpdateCmd := runCmds.InitVMUpdateCommand(vmCmd)
 	runCmds.InitVMUpdateTTL(vmUpdateCmd)
 
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -251,7 +251,7 @@ func Execute(rootCmd *cobra.Command, stdin io.Reader, stdout io.Writer, stderr i
 	runCmds.InitVMList(vmCmd)
 	runCmds.InitVMVersions(vmCmd)
 	runCmds.InitVMRemove(vmCmd)
-
+	runCmds.InitVMSSH(vmCmd)
 	vmUpdateCmd := runCmds.InitVMUpdateCommand(vmCmd)
 	runCmds.InitVMUpdateTTL(vmUpdateCmd)
 
@@ -265,7 +265,7 @@ func Execute(rootCmd *cobra.Command, stdin io.Reader, stdout io.Writer, stderr i
 	runCmds.InitNetworkList(networkCmd)
 	runCmds.InitNetworkRemove(networkCmd)
 	runCmds.InitNetworkJoin(networkCmd)
-	
+
 	networkUpdateCmd := runCmds.InitNetworkUpdateCommand(networkCmd)
 	runCmds.InitNetworkUpdateOutbound(networkUpdateCmd)
 
--- a/cli/cmd/vm_ssh.go
+++ b/cli/cmd/vm_ssh.go
@@ -78,15 +78,17 @@ func (r *runners) sshVM(cmd *cobra.Command, args []string) error {
 	runningVMs := filterVMsByStatus(vms, types.VMStatusRunning)
 	if len(runningVMs) == 0 {
 		// Show information about non-running VMs if they exist
-		activeVMs := filterActiveVMs(vms)
-		if len(activeVMs) == 0 {
+		nonTerminatedVMs := filterVMsByStatus(vms, "")
+		if len(nonTerminatedVMs) == 0 {
 			return errors.New("no active VMs found")
 		}
 
 		// List non-running VMs with their statuses
 		fmt.Println("No running VMs found. The following VMs are available but not running:")
-		for _, vm := range activeVMs {
-			fmt.Printf("  - %s (ID: %s, Status: %s)\n", vm.Name, vm.ID, vm.Status)
+		for _, vm := range nonTerminatedVMs {
+			if vm.Status != types.VMStatusTerminated {
+				fmt.Printf("  - %s (ID: %s, Status: %s)\n", vm.Name, vm.ID, vm.Status)
+			}
 		}
 		return fmt.Errorf("SSH connection requires a running VM. Please start a VM before connecting")
 	}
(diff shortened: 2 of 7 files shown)


jdewinne (commented):
Shouldn't there also be a check if the VM is in a `RUNNING` state before ssh'ing into it?

cli/cmd/vm_ssh.go
```diff
@@ -0,0 +1,110 @@
...
+
+	vms, err := r.kotsAPI.ListVMs(false, nil, nil)
+	if err != nil {
+		return errors.Wrap(err, "failed to list VMs")
+	}
+
+	// Filter out terminated VMs
+	activeVMs := filterActiveVMs(vms)
+	if len(activeVMs) == 0 {
+		return errors.New("no active VMs found")
```
squizzi: This should probably just be `no running VMs found` and then `filterActiveVMs` can just become `filterRunningVMs` based on Josh's feedback.

cli/cmd/vm_ssh.go
```diff
@@ -0,0 +1,110 @@
...
+	for _, vm := range vms {
+		if vm.Status != types.VMStatusTerminated {
+			activeVMs = append(activeVMs, vm)
+		}
+	}
+	return activeVMs
+}
+
+// selectVM prompts the user to select a VM from the list
+func selectVM(vms []*types.VM) (*types.VM, error) {
```
squizzi, line 131: This is nice and I think it'd be useful in other places like `cluster kubeconfig`, where we prompt for id when none is given.  Perhaps a good follow-up after this one.

cli/cmd/vm_ssh.go
```diff
@@ -0,0 +1,148 @@
...
+		if len(activeVMs) == 0 {
+			return errors.New("no active VMs found")
+		}
+
+		// List non-running VMs with their statuses
+		fmt.Println("No running VMs found. The following VMs are available but not running:")
+		for _, vm := range activeVMs {
+			fmt.Printf("  - %s (ID: %s, Status: %s)\n", vm.Name, vm.ID, vm.Status)
+		}
+		return fmt.Errorf("SSH connection requires a running VM. Please start a VM before connecting")
```
squizzi, line 93:
```suggestion
		return errors.New("SSH connection requires a running VM. Please start a VM before connecting")
```

cli/cmd/vm_ssh.go
```diff
@@ -0,0 +1,148 @@
...
+		// Show information about non-running VMs if they exist
+		activeVMs := filterActiveVMs(vms)
+		if len(activeVMs) == 0 {
+			return errors.New("no active VMs found")
+		}
+
+		// List non-running VMs with their statuses
+		fmt.Println("No running VMs found. The following VMs are available but not running:")
+		for _, vm := range activeVMs {
+			fmt.Printf("  - %s (ID: %s, Status: %s)\n", vm.Name, vm.ID, vm.Status)
```
squizzi: nit: Fancy bullet point or dash?  I like the bullet point.

```suggestion
			fmt.Printf("  • %s (ID: %s, Status: %s)\n", vm.Name, vm.ID, vm.Status)
```

cli/cmd/vm_ssh.go
```diff
@@ -0,0 +1,148 @@
...
+		}
+		return fmt.Errorf("SSH connection requires a running VM. Please start a VM before connecting")
+	}
+
+	// If only one running VM, use it directly
```
squizzi, line 96: So if I just run `replicated vm ssh` with no arguments and there's only one vm in my list that the filter catches I'll SSH into that one without any additional prompting?  I'm not sure I'm a fan of this.  It feels like odd UX to have a user expect a prompt when not specifying an ssh id and then only in specific situations they'll be ssh'd in.  I get what's trying to be accomplished here but I think it's just safer to make it consistent each time.  It also makes it difficult to understand if a user hasn't ran `r vm ls` and isn't aware that not all of the VMs they created are not yet `running`.

cli/cmd/vm_ssh.go
```diff
@@ -0,0 +1,144 @@
...
+	if len(args) == 1 {
+		vmID := args[0]
+
+		// Check VM status before connecting
+		vm, err := r.kotsAPI.GetVM(vmID)
+		if err != nil {
+			return errors.Wrap(err, "failed to get VM")
+		}
+
+		if vm.Status != types.VMStatusRunning {
```
squizzi, line 65: This check gets duplicated in https://github.com/replicatedhq/replicated/pull/512/files#diff-c575af5e04f2c78709e450752a0a55c0aa4b96e4507bea9013e98395f6be979aR24, it's probably not necessary in both places.  Perhaps simply `r.sshVM` to just get the flags from cobra and then `return r.kotsAPI.SSHIntoVM` and handle all of the validation there?

```go
func (r *runners) sshVM(cmd *cobra.Command, args []string) error {
	if err := r.initVMClient(); err != nil {
		return err
	}

	sshUser, _ := cmd.Flags().GetString("user")

        return r.kotsAPI.SSHIntoVM(vmID, sshUser)
}

File: pkg/kotsclient/vm_ssh.go [hunk 7]
```diff
@@ -0,0 +1,56 @@
...
+		return err
+	}
+
+	if vm.DirectSSHEndpoint == "" || vm.DirectSSHPort == 0 {
+		return fmt.Errorf("VM %s does not have SSH access configured", id)
+	}
+
+	// Check if VM is running
+	if vm.Status != types.VMStatusRunning {
+		return fmt.Errorf("VM %s is not running (current status: %s). SSH connection requires a running VM", id, vm.Status)
```
squizzi, line 24:
```suggestion
		return fmt.Errorf("VM: %s is not running (current status: %q): SSH connection requires a running VM", vmID, vm.Status)
```

pkg/kotsclient/vm_ssh.go
```diff
@@ -0,0 +1,56 @@
...
+	// Try to get the SSH user in order of precedence
+	sshUser := firstNonEmpty(
+		sshUserFlag,
+		os.Getenv("REPLICATED_SSH_USER"),
+		os.Getenv("GITHUB_ACTOR"),
+		os.Getenv("GITHUB_USER"),
+	)
+
+	sshEndpoint := vm.DirectSSHEndpoint
+	if sshUser != "" {
```
squizzi, line 36: What if `sshUser` _is_ empty?  We should probably error.  It might make sense for `firstNonEmpty` to just return `(string, error)` and `return "", errors.New("could not find non-empty string in given values")` or something when done.

hedge-sparrow: If a user is not provided and `sshUser` is empty, we fall back on the local ssh client's behavior for picking a user. 
by default it will use the local username unless otherwise configured in `~/.ssh/config`. I didn't want to block if no user was specified in the cases where people prefer to configure their user using the native ssh config.

cli/cmd/vm_ssh.go
hedge-sparrow: noted, one for another PR though

cli/cmd/root.go
```diff
@@ -251,7 +251,7 @@ func Execute(rootCmd *cobra.Command, stdin io.Reader, stdout io.Writer, stderr i
 	runCmds.InitVMList(vmCmd)
 	runCmds.InitVMVersions(vmCmd)
 	runCmds.InitVMRemove(vmCmd)
-
+	runCmds.InitVMSSH(vmCmd)
```
divolgin, line 254: Also need to call `runCmds.InitVMSSH(vmCmd)`.  Tried it locally, and it works.

hedge-sparrow: Oops yeh, missed `initVMSCP`

divolgin (approved):

PR #513, opened 2025-03-08
Performance Improvements for CLI Version Checking

marccampbell:
- Converted the blocking version update check to a non-blocking background operation
  - Eliminated the 3+ second delay on every CLI command
  - Added version cache invalidation to properly handle CLI upgrades
  - Improved handling for development and "unknown" versions
  - Added debug logging for version check diagnostics

  Background Version Checking

  - Moved the version check API call to a background goroutine that never blocks command execution
  - Used existing cache (at ~/.replicated/cache.json) to store and retrieve update information
  - Reduced the timeout for background checks from 3s to 1s
  - Added proper cache management with version-specific invalidation

  Development Version Handling

  - Added special handling for "unknown" and development versions
  - More gracefully handles semver parsing errors
  - Provides better update information for non-versioned builds

  Debug Mode

  - Added --debug flag and REPLICATED_DEBUG environment variable
  - Comprehensive debug logging throughout the version check process
  - Shows detailed timing for HTTP requests
  - Logs all cache reads/writes and version comparisons

  Performance Impact

  - Eliminated the ~3 second delay on every CLI command execution
  - Maintains the update notification functionality without affecting performance
  - Version command still performs a synchronous check to ensure accurate information

  Testing

  - Tested with production builds and development builds
  - Verified cached update detection works correctly
  - Confirmed cache is properly invalidated when CLI is upgraded
  - Validated debug mode provides appropriate diagnostics

### DIFF
--- a/cli/cmd/version.go
+++ b/cli/cmd/version.go
@@ -17,10 +17,40 @@ func Version() *cobra.Command {
 		Short: "Print the current version and exit",
 		Long:  `Print the current version and exit`,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			currentVersion := version.Version()
+				
+			// For version command, do a synchronous update check
+			updateChecker, err := version.NewUpdateChecker(currentVersion, "replicatedhq/replicated/cli") 
+			if err == nil {
+				// If we're in a development build or unknown version, still try to get latest
+				// version info but don't compare versions
+				updateInfo, err := updateChecker.GetUpdateInfo()
+				if err == nil && updateInfo != nil {
+					// Update the build info with latest version
+					version.SetUpdateInfo(updateInfo)
+					// Also save to cache for future commands
+					version.SaveUpdateCache(currentVersion, updateInfo)
+				}
+			}
+			
+			// Now get the (potentially updated) build info
 			build := version.GetBuild()
 
 			if !versionJson {
-				version.Print()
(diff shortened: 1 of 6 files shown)


cli/cmd/version.go
```diff
@@ -17,10 +17,40 @@ func Version() *cobra.Command {
...
+			}
+			
+			// Now get the (potentially updated) build info
 			build := version.GetBuild()
 
 			if !versionJson {
-				version.Print()
+				// Special handling for development/unknown version when printing
+				if currentVersion == "unknown" || currentVersion == "development" {
+					fmt.Printf("replicated version %s (development build)\n", currentVersion)
```
divolgin, line 42: this should go into stderr probably

pkg/version/updatecache.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get cache directory")
+	}
+
+	return filepath.Join(cacheDir, "cache.json"), nil
+}
+
+// LoadUpdateCache loads the update information from the cache file
+// Returns nil without error if the cache doesn't exist or is invalid
+func LoadUpdateCache() (*UpdateCache, error) {
```
divolgin, line 57: no point in having error in return

divolgin (approved):

PR #517, opened 2025-03-12
feature(cmx): Add 'network update policy'

squizzi:
Note: I'm opening this as a draft for now until we get the other backend additions in that will change outbound over to policy.  Once all of those pieces are in we can merge this and make the change over to `policy` instead of `outbound`.  This way we don't need to simultaneously support both `outbound` and `policy` side by side.

---

This PR adds support for configuring a network policy via the CLI and moves towards deprecating the `network update outbound` command, aliasing it to run `network update policy` commands when used.

Now, when a user wishes to configure a desired network policy they can run:

```
bin/replicated network update policy --id ce43abe6 --policy "airgap"
ID          NAME                           STATUS          CREATED                           EXPIRES                          POLICY
ce43abe6    angry_lumiere                  running         2025-03-12 15:37 PDT              2025-03-12 16:38 PDT             airgap
```

Outbound, now hidden still operates similarly and accepts `none` and `any` as arguments but now calls the underlying `UpdateNetworkPolicy` handler:

```
bin/replicated network update outbound --id ce43abe6 --outbound something
Note: 'replicated network update outbound' is deprecated. Use 'replicated network update policy' instead.
Error: outbound must be either 'none' or 'any'

bin/replicated network update outbound --id ce43abe6 --outbound 'none'
Note: 'replicated network update outbound' is deprecated. Use 'replicated network update policy' instead.
Updating policy to 'airgap' to match 'none' outbound setting
ID          NAME                           STATUS          CREATED                           EXPIRES                           POLICY
ce43abe6    angry_lumiere                  running         2025-03-12 15:37 PDT              2025-03-12 16:38 PDT              airgap

bin/replicated network update outbound --id ce43abe6 --outbound 'any'
Note: 'replicated network update outbound' is deprecated. Use 'replicated network update policy' instead.
Updating policy to 'open' to match 'any' outbound setting
ID          NAME                           STATUS          CREATED                           EXPIRES                           POLICY
ce43abe6    angry_lumiere                  running         2025-03-12 15:37 PDT              2025-03-12 16:38 PDT              open
```


squizzi:
Do we still want to support the `--network-policy` setting use case on `vm create` and `cluster create` or do we just want to continue the network UX where if you want to customize the `--network-policy` you first create a network.

cli/print/networks.go
```diff
@@ -10,18 +10,18 @@ import (
 )
 
 // Table formatting
-var networksTmplTableHeaderSrc = `ID	NAME	STATUS	CREATED	EXPIRES	OUTBOUND`
+var networksTmplTableHeaderSrc = `ID	NAME	STATUS	CREATED	EXPIRES	POLICY`
 var networksTmplTableRowSrc = `{{ range . -}}
-{{ .ID }}	{{ padding .Name 27	}}	{{ padding (printf "%s" .Status) 12 }}	{{ padding (printf "%s" (localeTime .CreatedAt)) 30 }}	{{if .ExpiresAt.IsZero}}{{ padding "-" 30 }}{{else}}{{ padding (printf "%s" (localeTime .ExpiresAt)) 30 }}{{end}}	{{if eq .Outbound ""}}{{ padding "-" 30 }}{{else}}{{ padding (printf "%s" (.Outbound)) 30 }}{{end}}
+{{ .ID }}	{{ padding .Name 27	}}	{{ padding (printf "%s" .Status) 12 }}	{{ padding (printf "%s" (localeTime .CreatedAt)) 30 }}	{{if .ExpiresAt.IsZero}}{{ padding "-" 30 }}{{else}}{{ padding (printf "%s" (localeTime .ExpiresAt)) 30 }}{{end}}	{{if eq .Policy ""}}{{ padding "-" 30 }}{{else}}{{ padding (printf "%s" (.Policy)) 30 }}{{end}}
```
jdewinne, line 15: Should "default" Policy be "Open" vs "-"?

jdewinne (approved):

cli/print/networks.go
squizzi: We set the default policy in the db as `"open"` so it shouldn't _ever_ be null, but we can set it for the sake of having it here.

jdewinne (approved):

PR #520, opened 2025-03-20
Consistent -o and remove duplicate default

ajp-io:
Consistently support `-o` as an alias for `--output`. Also remove duplicate mention of default.

Previously:
Screenshot 2025-03-20 at 1 31 06 PM

Now:
Screenshot 2025-03-20 at 1 31 42 PM


pandemicsyn (approved):
lgtm

PR #525, opened 2025-04-01
Pin action runner to `ubuntu-24.04` since 20.04 is EOLing

pandemicsyn:
Update the github actions to run on `ubuntu-24.04`

Fixes https://app.shortcut.com/replicated/story/121844/update-actions-using-the-ubuntu-20-04-lts-runner

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   make-unit-tests:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yaml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   make-unit-tests:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
divolgin, line 11: How about pinning it to 24.04 to reduce random breakages when upstream is updated?

PR #531, opened 2025-04-09
add `replicated vm ssh-endpoint` command to print the ssh endpoint of a vm

hedge-sparrow:
This enables users to use tools other than `ssh` or to construct more complicates commands with the endpoint. 

i.e: 
```bash
ssh $(replicated vm ssh-endpoint abdc123) 

scp ./file $(replicated vm ssh-endpoint abdc123):/some/path/

tar -c /path | ssh $(replicated vm ssh-endpoint abdc123) tar -xv
```


cli/cmd/vm_get_ssh_endpoint.go
```diff
@@ -0,0 +1,54 @@
...
+	githubUsername, err := r.kotsAPI.GetGitHubUsername()
+	if err != nil {
+		return errors.Wrap(err, "get github username")
+	}
+
+	// Format the SSH endpoint with username if available
+	if githubUsername != "" {
+		fmt.Printf("ssh://%s@%s:%d\n", githubUsername, vm.DirectSSHEndpoint, vm.DirectSSHPort)
+	} else {
+		fmt.Printf("ssh://%s:%d\n", vm.DirectSSHEndpoint, vm.DirectSSHPort)
```
squizzi, line 50: We should just return error here, if there's no github user affiliated with the account the user isn't going to be able to get into the box anyways because there's no key affiliated with them at that point.

We should error and point the user to updating their username to serve as a key on the Replicated portal.

squizzi (changes requested):
I think we should bring back the interactivity that was removed in #532, if I don't provide a VM ID I should still be able to select one from the list.  I liked that addition.  The difference being we'd just output the ssh URI from the selection rather than start a new SSH session.

cli/cmd/vm_get.go
```diff
@@ -0,0 +1,16 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func (r *runners) InitVMGetCommand(parent *cobra.Command) *cobra.Command {
```
squizzi, line 7: Is this getter neccessary?  The command results in `replicated vm ssh-endpoint`, but this is trying to place that underneath a sub-command like `replicated vm get ssh-endpoint`.  The getter isn't a bad idea, but currently with only one thing to get it seems like we should maybe add it later.  Based on how `replicated vm ssh-endpoint` works now we should just be able to remove this without much effect.

hedge-sparrow: oops this is a holdover I missed from the first agent's run...

hedge-sparrow:
> I think we should bring back the interactivity that was removed in #532, if I don't provide a VM ID I should still be able to select one from the list. I liked that addition. The difference being we'd just output the ssh URI from the selection rather than start a new SSH session.

The problem with having interactivity in something we expect to run in subshells is that it doesn't always work the way we expect, things run inside bash's `$()` or fish's `()` don't behave the same way as if run directly from a tty

squizzi:
> The problem with having interactivity in something we expect to run in subshells is that it doesn't always work the way we expect, things run inside bash's $() or fish's () don't behave the same way as if run directly from a tty

Maybe a happy middle ground would be to print a `vm ls` of running status VMs in the case where an ID is fudged or the command is attempted against an ID that is not yet running.

I dunno, I see a benefit to keeping `vm ssh` around if thats the case -- which I guess we could discuss in greater detail on that PR because I find it moderately annoying to do a `r vm ls` and then make sure the VM is running and then `ssh $(replicated vm ssh-endpoint <id>)` 

I'm going to approve this and then we can discuss the potential of keeping `vm ssh` around for _just the interactive `ssh`_ use case in the other PR.

squizzi (approved):

PR #532, opened 2025-04-10
Remove `replicated vm ssh`

hedge-sparrow:
This removes the `ssh` wrapper functionality


squizzi (dismissed):

squizzi:
We should try adding the `--no-heading` option to support the use of `fzf` to replicated some of the interactivity that's lost, we can document that approach to users as well.

AmberAlston (dismissed):

squizzi (approved):

PR #554, opened 2025-07-07
upgrades helm dep

staciegravy:
Updates the Helm dependency and resolves the following stories and dependabot alerts. 

Notable mentions: 
This upgrade required an upgrade to Go 1.24+, which is more strict around string interpolation, so I had to fix a couple spots. 

* https://app.shortcut.com/replicated/story/126206/helm-allows-a-specially-crafted-chart-archive-to-cause-out-of-memory-termination-airgap-replicated
* https://app.shortcut.com/replicated/story/126207/helm-allows-a-specially-crafted-json-schema-to-cause-a-stack-overflow-airgap-replicated

Resolves: https://github.com/replicatedhq/replicated/security/dependabot/59
Resolves: https://github.com/replicatedhq/replicated/security/dependabot/60

### DIFF
--- a/dagger/docs.go
+++ b/dagger/docs.go
@@ -66,7 +66,7 @@ func (r *Replicated) GenerateDocs(
 
 	// generate the docs from this current commit
 	docs := dag.Container().
-		From("golang:1.23").
+		From("golang:1.24").
 		WithMountedDirectory("/go/src/github.com/replicatedhq/replicated", source).
 		WithWorkdir("/go/src/github.com/replicatedhq/replicated").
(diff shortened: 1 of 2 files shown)


pandemicsyn (changes requested):
@staciegravy mind bump'ing the Go version we use in dagger?

https://github.com/replicatedhq/replicated/blob/62917d9a838e93cf0aa7ae474fdb310b8d407f12/dagger/release.go#L114

https://github.com/replicatedhq/replicated/blob/62917d9a838e93cf0aa7ae474fdb310b8d407f12/dagger/functionality.go#L19

PR #573, opened 2025-08-21
chore: unhide network commands

mikhailswift:
Unhides the network root command and ls subcommand from the CLI usage output. This will make them more easily discoverable as well as enable doc generation for them.


cli/cmd/network_join.go
```diff
@@ -6,10 +6,11 @@ import (
...
 	cmd := &cobra.Command{
-		Use:   "join",
-		Short: "Join a test network",
-		Long:  ``,
-		RunE:  r.joinNetwork,
+		Use:    "join",
+		Short:  "Join a test network",
+		Long:   ``,
+		RunE:   r.joinNetwork,
+		Hidden: true,
```
jdewinne, line 13: Good point. But what does this command even do? 🤔

jdewinne (approved):
Just one nit. I have no idea what network join actually does or should do.

cli/cmd/network.go
```diff
@@ -11,10 +12,13 @@ import (
...
-		Use:    "network",
-		Short:  "Manage test networks for VMs and Clusters",
-		Long:   ``,
-		Hidden: true,
+		Use:   "network",
+		Short: "Manage test networks for VMs and Clusters",
+		Long: `The 'network' command allows you to manage and interact with networks used for testing purposes.
+With this command you can list the networks in use by VMs and clusters.`,
+		Example: `# List all networks
+replicated network ls`,
```
squizzi, line 20: `network ls` columns are going to show `REPORTING` and `POLICY` even though those are airgap features.  In the past this wasn't a major issue for Amber (like on the CMX history page), but it might be worth double checking that that's ok with product.

squizzi (approved):
lgtm other then the `network ls` columns potentially containing alpha features

PR #586, opened 2025-09-19
Add source and destination pod

jdewinne:
Example output
```
{"timestamp":"2025-09-19T01:27:27.671401693Z","srcIp":"10.42.1.2","dstIp":"162.159.138.43","srcPort":11984,"dstPort":80,"sourcePod":"default/netshoot-l9574","proto":"TCP","likelyService":"http"}
{"timestamp":"2025-09-19T01:27:27.682874197Z","srcIp":"10.42.1.2","dstIp":"162.159.138.43","srcPort":11984,"dstPort":80,"sourcePod":"default/netshoot-l9574","proto":"TCP","comm":"containerd","pid":2179,"likelyService":"http"}
```

### DIFF
--- a/pkg/types/network.go
+++ b/pkg/types/network.go
@@ -41,14 +41,16 @@ type NetworkReport struct {
 }
 
 type NetworkEventData struct {
-	Timestamp     string `json:"timestamp"`
-	SrcIP         string `json:"srcIp"`
-	DstIP         string `json:"dstIp"`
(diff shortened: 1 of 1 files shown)


squizzi:
I don't think we'll ever use destination pod because we only cover traffic that is leaving the cluster, not intracluster traffic 🤔 , but if we ever decide to provide more network reporting granularity it doesn't hurt to have it.

PR #691, opened 2026-04-15
pass through RBAC errors in vm create

laverya:
```
./bin/replicated vm create --distribution ubuntu --version 22.04
Error: access to "kots/vm/create" is denied
```

### DIFF
--- a/cli/cmd/vm_create.go
+++ b/cli/cmd/vm_create.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/moby/moby/pkg/namesgenerator"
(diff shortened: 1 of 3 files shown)


cli/cmd/vm_create.go
```diff
@@ -175,6 +179,11 @@ func (r *runners) createAndWaitForVM(opts kotsclient.CreateVMOpts) ([]*types.VM,
 	return vms, nil
 }
 
+func isRBACDeniedError(err error) bool {
+	message := strings.TrimSpace(strings.ToLower(err.Error()))
+	return strings.HasPrefix(message, "access to ") && strings.HasSuffix(message, " is denied")
+}
+
```
Justiceleeg, line 186: is there no better way to decide if its an RBAC error than by string matching?

laverya: RBAC errors are unfortunately just a 403 and a string
it's not json or anything

Justiceleeg (approved):

cli/cmd/vm_create.go
laverya: The literal API response is `access to "kots/vm/create" is denied` lol

PR #715, opened 2026-05-05
feat: ability to create releases without uploading them

banjoh:
## Summary

Adds two new flags to `replicated release create` so a release can be built locally and inspected (or fed into another command) without uploading it to the vendor portal:

- `--no-upload`: build the release locally and skip the API upload (and any promotion). Rejects `--promote` / `--ensure-channel` / `--required`.
- `--output-dir <path>`: stage the release artifacts (packaged charts + collected manifests for the `.replicated` config flow, or the raw release JSON payload for `--yaml`/`--yaml-file`/`--chart`
flows) at the given path and preserve them after the command completes.

The two flags are orthogonal: typical usage is `--no-upload --output-dir ./out` to stage artifacts for inspection, while `--output-dir` on its own keeps a copy of what was uploaded.

Online mode is still required — the command continues to resolve the app type via the API even with `--no-upload`. Going fully offline is out of scope here.

Also drops an unused `getToolVersion` helper in `cli/cmd/lint.go` flagged while editing nearby code.

Refs: [sc-137535](https://app.shortcut.com/replicated/story/137535)


cli/cmd/lint.go
```diff
@@ -21,16 +21,6 @@ import (
 // release-validation-v2 feature flag. The runLint function below is still used
 // internally by the release lint command.
 
-// getToolVersion extracts a tool version from config, defaulting to "latest" if not found.
-func getToolVersion(config *tools.Config, tool string) string {
```
banjoh, line 25: This was an unused function I cleaned up

laverya (dismissed):

laverya (approved):

PR #718, opened 2026-05-06
feat(lint): discover helm charts from EC Config manifests

emosbaugh:
## Summary

Add support for discovering HelmChart declarations from `embeddedcluster.replicated.com/v1beta1` Config manifest `extensions.helmCharts`.

This allows validation to properly match EC Config helm charts against local chart archives, preventing false-positive warnings for charts declared in EC Config manifests.

## Changes

- `DiscoverECConfigHelmCharts()` scans manifest globs for EC Config documents and extracts `spec.extensions.helmCharts[].chart` entries
- `DiscoverHelmChartManifests()` now merges both kots `HelmChart` manifests and EC Config helm chart declarations
- Duplicate detection works across both source types
- Unit tests and integration test scenarios added for valid matches and missing archives

### DIFF
--- a/pkg/lint2/helmchart.go
+++ b/pkg/lint2/helmchart.go
@@ -121,6 +121,22 @@ func DiscoverHelmChartManifests(manifestGlobs []string) (map[string]*HelmChartMa
 		}
 	}
 
+	// Discover EC Config helm charts and merge
+	ecHelmCharts, err := DiscoverECConfigHelmCharts(manifestGlobs)
+	if err != nil {
+		return nil, err
+	}
+	for key, manifest := range ecHelmCharts {
+		if existing, found := helmCharts[key]; found {
+			return nil, &DuplicateHelmChartError{
+				ChartKey:   key,
+				FirstFile:  existing.FilePath,
+				SecondFile: manifest.FilePath,
+			}
+		}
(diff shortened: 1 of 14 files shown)


emosbaugh:
[sc-137591]

kriscoleman (dismissed):
:shipit:

pkg/lint2/helmchart.go
```diff
@@ -136,6 +152,71 @@ func isHelmChartManifest(path string) (bool, error) {
...
+			if isHiddenPath(path) {
+				continue
+			}
+			if seenFiles[path] {
+				continue
+			}
+			seenFiles[path] = true
+
+			// Quick kind check before parsing
+			isConfig, err := hasKind(path, "Config")
```
sgalsaleh: there's multiple `Config` kinds with different api versions (e.g. KOTS's config, ec's Config, etc...). i guess the below parse continues on error so that might be fine with the current implementation you have but might cause issues later on.

sgalsaleh (dismissed):

pkg/lint2/helmchart.go
emosbaugh: added hasAPIVersionKind function

sgalsaleh (approved):

PR #723, opened 2026-05-17
get latest release should not include demoted

marccampbell:
problem:  GetCurrentChannelRelease defines current as the release with the highest channel sequence in both the embedded-release and fallback paths. ChannelRelease explicitly carries demotion state, but neither path excludes IsDemoted releases. If the latest promoted release was demoted, commands that ask for the current channel release can operate on the demoted release instead of the active one, returning stale or intentionally removed release data such as image lists and proxy metadata.

solution:
When selecting the current release, skip releases with IsDemoted set. If all releases are demoted, return a clear no active release error. Keep the same filtering in both the kotsChannel.Releases path and the ListChannelReleases fallback path.

### DIFF
--- a/client/channel.go
+++ b/client/channel.go
@@ -230,7 +230,7 @@ func (c *Client) GetCurrentChannelRelease(appID string, appType string, channelI
 
 		// If the channel has releases data, find the current one
 		if len(kotsChannel.Releases) > 0 {
-			currentRelease := currentChannelRelease(kotsChannel.Releases)
+			currentRelease := currentChannelRelease(kotsChannel.Releases, kotsChannel.ChannelSequence)
 			if currentRelease != nil {
 				proxyDomain := currentRelease.ProxyRegistryDomain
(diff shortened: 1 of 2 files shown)


NoaheCampbell (dismissed):

client/channel.go
```diff
@@ -300,6 +294,32 @@ func (c *Client) GetCurrentChannelRelease(appID string, appType string, channelI
 	return nil, "", errors.Errorf("unknown app type %q", appType)
 }
 
+func currentChannelRelease(releases []types.ChannelRelease) *types.ChannelRelease {
```
divolgin: This doesn't account for semver sorting, but I also believe the API should be sorting releases already. There should also be current release fields set on the channel response: https://replicated-vendor-api.readme.io/reference/getchannelproperties-1  (I see `latestChannelSequence`, `releaseSequence` and `currentVersion` fields)

divolgin (approved):

PR #724, opened 2026-05-17
properly handle app type in api

marccampbell:
problem: The exported aggregate client exposes GetApp, CreateApp, and DeleteApp as successful operations, but they either return a nil result with nil error or a nil error without calling either backend. The context platform client has real CreateApp/DeleteApp implementations, so callers using the higher-level client can believe an app was fetched, created, or deleted when no API request happened. DeleteApp is especially risky because it silently reports success for a no-op.

solution: Implement these methods by delegating to the appropriate platform/KOTS client based on app type or input options. If this abstraction cannot support an operation, return a non-nil unsupported-operation error rather than nil success.

### DIFF
--- a/client/app.go
+++ b/client/app.go
@@ -3,6 +3,9 @@ package client
 import (
 	"context"
 
+	"github.com/pkg/errors"
+	"github.com/replicatedhq/replicated/pkg/kotsclient"
+	"github.com/replicatedhq/replicated/pkg/platformclient"
 	"github.com/replicatedhq/replicated/pkg/types"
 )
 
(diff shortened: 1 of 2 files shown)


NoaheCampbell (dismissed):

client/app.go
```diff
@@ -49,13 +52,47 @@ func (c *Client) ListApps(excludeChannels bool) ([]types.AppAndChannels, error)
...
+	if err != nil {
+		return nil, err
+	}
+	return app, nil
 }
 
 func (c *Client) CreateApp(appOptions interface{}) (interface{}, error) {
-	return nil, nil
+	switch opts := appOptions.(type) {
+	case string:
```
divolgin: I'm not really sure what this case means.  The `DeleteApp` function handles this differently.  But also neither CreateApp nor DeleteApp are used anywhere outside of tests.

divolgin (approved):

PR #740, opened 2026-05-21
feat(cli): add global --output persistent flag

nvanthao:
## Summary

Introduce a global `--output` (`-o`) persistent flag on the root command that all subcommands inherit, replacing ~78 local `--output` flag registrations. Also adds support for a new `REPLICATED_OUTPUT` environment variable for setting a default output format.

## Resolution Order

1. Explicit `--output` flag on the command line
2. `REPLICATED_OUTPUT` environment variable
3. Default: `"table"`

### Core Infrastructure
- Added `resolveOutputFormat()` to runners with resolution order: explicit flag → env var → "table" default
- Registered global persistent flag on root command
- Wired resolution into `preRunSetupAPIs` and `prerunCommand`
- Added `Proxy: http.ProxyFromEnvironment` to the platform HTTP client transport so the CLI now respects `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` environment variables

### Command Refactoring
- **Migrated `version` command** from `--json` bool flag to global `--output` (legacy `--json` preserved but deprecated)
- **Refactored 17 commands** that declared local `outputFormat` variables to use `r.outputFormat` directly:
  - app (ls, create, rm, hostname ls)
  - customer (ls, create, inspect)
  - policy (ls, get, create, update)
  - default (show, set)
  - enterprise-portal (status get/update, invite, user ls)
- **Removed redundant local flags** from ~50 commands already binding to `r.outputFormat`
- **Refactored notification.go** — all 10 subcommands now use `r.outputFormat`
- **Refactored cluster addon commands** — `cluster_addon_ls.go` and `cluster_addon_create_objectstore.go`
- **Added JSON output support** to previously table-only commands:
  - `channel adoption` — full JSON marshaling of `ChannelAdoption` struct
  - `channel counts` — full JSON marshaling of `LicenseCounts` struct
  - `release promote` — structured JSON output with `channel`, `release_sequence`, and `version_label`
  - `release image ls` — `print.ChannelImages` now supports JSON format
- **Added `SilenceUsage: true`** to hidden commands `channel adoption` and `channel counts`

### Clean Up
- Removed `clusterAddonCreateObjectStoreOutput` from `runnerArgs`

### Tests
- Added `cli/cmd/output_test.go` with 3 unit tests verifying resolution priority
- Added `cli/print/channel_adoption_test.go` (5 tests)
- Added `cli/print/channel_images_test.go` (4 tests)
- Added `cli/print/channel_license_counts_test.go` (5 tests)

### Documentation
- Added `docs/plans/global-output-flag.md` containing the full design and implementation plan

## Exclusions (Intentionally Left Unchanged)
- `customer download-license --output` — still a **file path** flag
- `login`, `logout`, `completion`, `config` — no output formatting

## Verification
- [x] `go build ./...` — zero errors
- [x] `go test ./cli/cmd/...` — all pass
- [x] Global `--output` appears in `replicated --help`
- [x] `REPLICATED_OUTPUT=json` works without explicit `--output json`
- [x] Explicit `--output wide` still overrides env var
- [x] `customer download-license --output` still works as a file path flag


cli/cmd/root.go
```diff
@@ -436,6 +443,7 @@ func Execute(rootCmd *cobra.Command, stdin io.Reader, stdout io.Writer, stderr i
 	}
 
 	prerunCommand := func(cmd *cobra.Command, args []string) (err error) {
+		runCmds.resolveOutputFormat(cmd)
```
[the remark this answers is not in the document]
nvanthao: preRunSetupAPIs is called separately as well

docs/plans/global-output-flag.md
marccampbell, line 1: this file is likely accidentally committed, lets not commit it

pkg/platformclient/client.go
```diff
@@ -29,6 +29,7 @@ var (
 // This is a singleton that's reused for all requests to avoid leaking connections.
 var httpClient = &http.Client{
 	Transport: &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
```
xavpaice, line 32: nit: is this related to the change?  Just want to understand the reason, rather than remove the change.

xavpaice (commented):
- Clean consolidation of output formatting logic. Using r.outputFormat everywhere and resolving it via resolveOutputFormat() in the two persistent pre-run hooks is the right approach.
- Explicit flag correctly wins over env var, and env var correctly wins over default. Tests cover this.
- JSON support added for channel adoption, channel counts, release image ls, and release promote is consistent with the rest of the CLI.
- version --json deprecation path is handled gracefully.
- Good test coverage for the new print functions.

Issues

cli/cmd/network_update.go:51 still registers its own local --output flag:
`cmd.Flags().StringVar(&r.outputFormat, "output", "table", "The output format to use. One of: json|table|wide")`, was this one missed?

cli/cmd/customer_download_license.go:42 uses -o / --output for a file path:
`cmd.Flags().StringVarP(&output, "output", "o", "-", "Path to output license to. Defaults to stdout")`

The local flag shadows the global format flag, so the command still works for its intended purpose. However, this means --output json on this command writes the license to a file literally named json instead of formatting as JSON.  Rather than change the command in this PR, let's open a followup issue for this inconsistency.  We don't want to make a breaking change here, but a future PR could add a `--output-file` or `--path` for the file output and consider if the json output is useful for this command.

I haven't yet run each command to test the look of the json output.  When we discuss 'wire format' in the Shortcut, are we suggesting that the entire API json response should be posted?  Is that huge and problematic, or deliberate?

pkg/platformclient/client.go
nvanthao: I saw this during testing the changes, the CLI currently not respecting HTTP proxy env vars (HTTP_PROXY, HTTPS_PROXY, NO_PROXY), adding this here so that the CLI will respect the proxy settings.

nvanthao:
@xavpaice 

- I have updated global flag for network_update
- Follow-up issue opened for download license https://github.com/replicatedhq/replicated/issues/748
- Yes I believe the `wire format` is to print out entire API json with exeception `release promote` and `release image ls`, for the other 3 commands `channel counts, channel releases, channel adoption`, these are for heritage app

xavpaice (approved):

PR #745, opened 2026-05-21
fix(cli): customer inspect does not surface isHelmInstallEnabled field

kriscoleman:
## Summary

- Add `IsHelmInstallEnabled` field with `json:"isHelmInstallEnabled"` tag to the `Customer` struct in `pkg/types/customer.go`
- The Vendor API returns `isHelmInstallEnabled` in customer responses, but the CLI was only mapping to the retired `isHelmVmDownloadEnabled` tag — causing `customer inspect --output json` to silently drop the value

## Root Cause

The `Customer` struct had no field for `isHelmInstallEnabled`. Go's JSON unmarshaler silently ignores unknown keys, so the API response value was lost.

## Test Plan

- [x] `replicated customer create --helm-install ...` followed by `replicated customer inspect --output json` now includes `isHelmInstallEnabled: true`
- [x] Existing `isHelmVmDownloadEnabled` field preserved for backwards compatibility
- [x] `go build ./...` passes
- [x] `go test ./pkg/types/... ./cli/print/...` passes

Fixes: [sc-137962]

🤖 Generated with [Claude Code](https://claude.com/claude-code)


xavpaice (dismissed):
LGTM, appreciate adding the tests.  We add tests here specifically for IsHelmInstallEnabled and a more broad test also, are the specific tests redundant?

kriscoleman:
> LGTM, appreciate adding the tests. We add tests here specifically for IsHelmInstallEnabled and a more broad test also, are the specific tests redundant?

Yea, I was trying to be thorough but it's probably a bit much, they are essentially testing the same thing, I can clean that up

kriscoleman:
Cleaned up in dd5f95f0 — removed the specific `IsHelmInstallEnabled` and `IsHelmVMDownloadEnabled` tests, kept the `BothHelmFields` and `FullAPIPayload` tests which cover everything without the redundancy. Thanks for the callout!

xavpaice (approved):

PR #763, opened 2026-07-27
fix: align DoJSONWithoutUnmarshal with DoJSON contract

marccampbell:
## Why this matters

The `replicated api get|post|put|patch` commands go through `DoJSONWithoutUnmarshal`, a weaker HTTP path that diverged from the main `DoJSON` client used by the rest of the CLI.

That meant those commands:

- **Could not be canceled** — requests were built with `http.NewRequest` (no context). Ctrl+C / signal cancellation and caller deadlines never reached the in-flight HTTP call. A hung API server could block the CLI indefinitely (the shared client has no overall timeout).
- **Did not identify the CLI to the API** — no `User-Agent` header (`Replicated/<version>`), so vendor-side logs and support could not tell which CLI version made the request.
- **Did not mark CI / GitHub Actions traffic** — missing `X-Replicated-CI` and related GitHub Actions headers that `DoJSON` sets when running in automation, so CMX and other backend linking that depends on those headers never applied to raw `api` calls.
- **Lost structured error handling** — non-2xx responses were generic string errors. Callers could not use `errors.Is(..., ErrNotFound)` for 404s, and 403s did not parse into `ForbiddenError` with the server message.

Most other kotsclient/platformclient methods already use `DoJSON` and get the full contract. The `api` helpers were the odd path out.

## What changed

- Align `DoJSONWithoutUnmarshal` with `DoJSON`: context via `NewRequestWithContext`, shared header enrichment (`setCommonHeaders`), and consistent 404 / 403 / `APIError` handling.
- Thread context through kotsclient `Get`/`Post`/`Put`/`Patch`.
- Pass `cmd.Context()` from the `api` CLI commands so cancellation actually works.

## Test plan

- [x] `go test ./pkg/platformclient/ ./pkg/kotsclient/`
- [x] `go build ./cli/... ./pkg/platformclient/ ./pkg/kotsclient/`
- [ ] Manually: `replicated api get /v3/apps` still returns JSON
- [ ] Manually: Ctrl+C interrupts an in-flight `api get` against a slow/hung endpoint

### DIFF
--- a/pkg/platformclient/client.go
+++ b/pkg/platformclient/client.go
@@ -129,7 +129,7 @@ func (c *HTTPClient) GetOrigin() string {
 	return c.apiOrigin
 }
 
-func (c *HTTPClient) DoJSONWithoutUnmarshal(method string, path string, reqBody string) ([]byte, error) {
+func (c *HTTPClient) DoJSONWithoutUnmarshal(ctx context.Context, method string, path string, reqBody string) ([]byte, error) {
 	endpoint := fmt.Sprintf("%s%s", c.apiOrigin, path)
 	var buf *bytes.Buffer
 	if reqBody != "" {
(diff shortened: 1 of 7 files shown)


pkg/platformclient/client.go
```diff
@@ -129,22 +129,26 @@ func (c *HTTPClient) GetOrigin() string {
...
+func (c *HTTPClient) DoJSONWithoutUnmarshal(ctx context.Context, method string, path string, reqBody string) ([]byte, error) {
 	endpoint := fmt.Sprintf("%s%s", c.apiOrigin, path)
 	var buf *bytes.Buffer
 	if reqBody != "" {
 		buf = bytes.NewBuffer([]byte(reqBody))
 	} else {
 		buf = &bytes.Buffer{}
 	}
-	req, err := http.NewRequest(method, endpoint, buf)
+	if ctx == nil {
```
divolgin, line 140: Odd pattern. I can't think of a reason why ctx would ever be nil; the caller should always pass one in.

divolgin: But I see `DoJSON` does the same thing. So this was probably just copied to follow the same pattern.

divolgin (approved):

PR #770, opened 2026-07-27
fix: bound network response body reads

marccampbell:
## Why this matters

Some production paths buffered entire network response bodies with unbounded `io.ReadAll`. For **small, fixed-shape** payloads (token JSON, CMX metadata, checksum text, OCI manifests), a modest ceiling is a reasonable guard without changing normal product behavior.

## Review feedback

Vendor API responses in `pkg/platformclient` are **not** limited. Per review: this is a short-lived CLI against a trusted Vendor API; a hard cap can turn a successful large release payload into a client error without meaningful security benefit. That change was dropped.

## What remains bounded

| Path | Limit | Rationale |
|------|-------|-----------|
| Login token exchange | 1 MiB | Token JSON only |
| CMX MMDS metadata + token | 1 MiB | Small credential payloads |
| Helm/troubleshoot checksum files | 1 MiB | Text checksum lists |
| OCI model manifest | 16 MiB | Manifest JSON, not layers |

## Explicit non-goals

- No limit on Vendor API client responses (`platformclient`)
- No limit on CLI stdin release YAML
- No limit on tool binary/archive downloads

## Test plan

- [x] `go test ./pkg/platformclient/ ./pkg/credentials/ ./pkg/tools/ ./cli/cmd/`
- [x] Platform client uses unbounded `io.ReadAll` again (no size cap)

### DIFF
--- a/pkg/platformclient/client.go
+++ b/pkg/platformclient/client.go
@@ -19,10 +19,6 @@ import (
 
 const apiOrigin = "https://api.replicated.com/vendor"
 
-// maxResponseBody is a generous ceiling for Vendor API response bodies.
-// Large multi-document releases must still fit; this only bounds unbounded allocation.
-const maxResponseBody = 100 << 20 // 100 MiB
-
 var (
 	ErrForbidden = errors.New("the action is not allowed for the current user or team")
 )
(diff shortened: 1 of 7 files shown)


pkg/platformclient/client.go
```diff
@@ -242,6 +246,17 @@ func (c *HTTPClient) DoJSON(ctx context.Context, method string, path string, suc
 	return nil
 }
 
+func readAllLimited(r io.Reader, limit int64) ([]byte, error) {
```
divolgin: This seems unnecessary. This can potentially turn a successful API request into an error causing the client to retry. But also this is a CLI tool, not a long running service, so response size should not be a major issue.

divolgin (approved):

PR #771, opened 2026-07-28
fix: search channel releases so release image ls --version finds releases efficiently

xavpaice:
When running `replicated release image ls --channel ... --version ...`, the CLI was only fetching the first page of channel releases (default page size of 20) and then filtering client-side by semver. This caused versions beyond the first page to fail with `no release found with version ... in channel`, even though the release existed and was visible in the UI.

This change:

1. Uses the API's server-side `versionLabel` filter when a specific version is requested, so only matching releases are returned instead of paginating through the entire channel history.
2. Keeps the previous pagination path as a fallback for edge cases where the version label differs from the semver.
3. Keeps the broader fix to `ListChannelReleases` so it paginates through all pages (used by the fallback and by `GetCurrentChannelRelease`).
4. Adds tests covering pagination behavior and the new `versionLabel` filter.

Fixes the reported issue where `2.17.12` in `itrs-analytics/unstable` could not be found via the CLI but was visible via the API on page 1.


cli/cmd/release_image_ls.go
```diff
@@ -173,6 +169,24 @@ func cleanImageName(image string, proxyRegistryDomain string) string {
 	return cleaned
 }
 
+// findReleaseByVersion tries the API's versionLabel filter first, and falls back
+// to listing all releases when the filter errors or returns no semver match.
+func (r *runners) findReleaseByVersion(channelID string, version string) (*types.ChannelRelease, error) {
+	filteredReleases, err := r.api.ListChannelReleasesByVersion(r.appID, r.appType, channelID, version, r.args.releaseImageLSIncludeInstallerImages)
+	if err == nil {
+		if release, err := findTargetRelease(filteredReleases, version); err == nil {
```
divolgin: findTargetRelease should be deleted and not used anywhere.

- `ListChannelReleasesByVersion` returns releases sorted by channel sequence and `findTargetRelease` will always return the first item on the list.
- if `ListChannelReleasesByVersion` does not match a release, `ListChannelReleases` will not return any matching releases either and `findTargetRelease` will always return ~nil~ error.

There is no need for a fallback here since version string is never empty. But also `ListChannelReleases` will be very slow when there are a lot of releases on a channel.

divolgin (changes requested):

pkg/kotsclient/channel_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/replicatedhq/replicated/pkg/platformclient"
+	"github.com/replicatedhq/replicated/pkg/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestListChannelReleases_Pagination(t *testing.T) {
```
divolgin: There are tests for this already https://github.com/replicatedhq/replicated/blob/39f852b22ee9bb028b878bc03d93a5261d2b24ef/pkg/integration/channel_test.go#L67

divolgin (approved):

